PR #47, opened 2015-10-26
Escape shell args to prevent spaces etc.

kelunik:


lib/WatcherProcess.php
```diff
@@ -248,10 +249,11 @@ private function onDeadIpcClient(string $readWatcherId, $ipcClient) {
...
         $parts[] = \PHP_BINARY;
         if ($ini = \get_cfg_var("cfg_file_path")) {
-            $parts[] = "-c \"{$ini}\"";
+            $ini = escapeshellarg($ini);
+            $parts[] = "-c {$ini}";
         }
         $parts[] = "-d zend.assertions=" . ini_get("zend.assertions");
-        $parts[] = __DIR__ . "/../bin/aerys-worker";
+        $parts[] = escapeshellarg(__DIR__ . "/../bin/aerys-worker");
         $parts[] = "-i {$this->ipcServerUri}";
```
staabm: what about this var?

lib/WatcherProcess.php
```diff
@@ -248,10 +249,11 @@ private function onDeadIpcClient(string $readWatcherId, $ipcClient) {
...
         if ($console->isArgDefined("config")) {
             $parts[] = "-c " . $console->getArg("config");
```
staabm: are those console-args "secure-by-default"?

lib/WatcherProcess.php
kelunik: Shouldn't be a problem, but may be a problem when switched to unix domain sockets and spaces in file paths.

lib/WatcherProcess.php
kelunik: Should be, but anyway. Added escaping for all args.

staabm:
after all its also a security fix

kelunik:
We need a security contact address for real security issues.

staabm:
> We need a security contact address for real security issues.

agree. this should also be documented on the website and github repos on how-to report security related issues.

kelunik:
See #48.

bwoebi:
Merged as 7bec44a with a few improvements

PR #50, opened 2015-11-01
Add static file root instructions to readme

rdlowrey:


camspiers:
It would be great to provide a short example:


## Static File Serving

To start a static file server simply pass a root handler as part of your config file. For example to serve a "public" directory in your app:

```php
(new Aerys\Host)
  ->expose("*", 1337)
  ->use(Aerys\root(__DIR__."/public"))
;
```

kelunik:
See https://github.com/amphp/aerys/commit/9e2ff723e898c0f35ea521a388638ce48947ec4d.

camspiers:
:+1:

PR #52, opened 2015-11-03
Align log messages for better readability

kelunik:


staabm:
Maybe this should only be done in non production mode?

bwoebi:
Is that really nicer? I don't like it that much.
```
[21:37:23] info      Using config file found at /Users/Bob/Aerys/test.php
[21:37:23] info      Listening on tcp://192.168.1.102:1337
[21:37:23] info      Listening on tcp://127.0.0.1:1337
```

That looks like bogus space… :-/

kelunik:
Yes, because `info` and `debug` are short, but it's really not that nice to read then it's always off by one character. It's something we need to discuss, that's why it's a PR, not a direct commit.

bwoebi:
Yeah, but 5 spaces is just too much … I think this isn't really that nice.

kelunik:
Maybe just shorten everything else to 4 or 5 chars?

bwoebi:
Not sure; it might be unusual to users. But anyway, from my experience, most logging does not do any alignment and so we shouldn't either.

"emerg", "alert", "crit", "error", "warn", "note", "info", "debug"

should be fine, with at most 2 spaces.

bwoebi:
The only issue is that it won't be PSR compatible then; see PsrLogLevel constants.

PR #56, opened 2015-11-04
Clean up buggy code

tpunt:
1. `$name` of `__construct()` will always be a string, so the first branch of the ternary will always be taken.
2. The IF statement comparing that `$this->name` does not equal `null` will always take the first branch since `$this->name` will always be a string (due to the assignment on first line of method).
3. `date_default_timezone_get()` will always have a value, so no need to default it to UTC

### DIFF
--- a/lib/Vhost.php
+++ b/lib/Vhost.php
@@ -50,7 +50,7 @@ class Vhost {
     ];
 
     public function __construct(string $name, array $interfaces, callable $application, array $filters) {
-        $this->name = isset($name) ? strtolower($name) : "";
+        $this->name = strtolower($name);
         if (!$interfaces) {
             throw new \InvalidArgumentException(
                 "At least one interface must be passed, an empty interfaces array is not allowed"
@@ -66,7 +66,7 @@ class Vhost {
             $this->tlsDefaults["alpn_protocols"] = "h2";
         }
 
-        if ($this->name !== null) {
+        if ($this->name !== '') {
             $addresses = [$this->name];
         } else {
             $addresses = array_unique(array_column($interfaces, 0));
@@ -248,7 +248,7 @@ class Vhost {
         }
 
         if (time() > $cert['validTo_time_t']) {
-            date_default_timezone_set(@date_default_timezone_get() ?: 'UTC');
+            date_default_timezone_set(@date_default_timezone_get());
             $expiration = date('Y-m-d', $cert['validTo_time_t']);
             trigger_error(
                 "TLS certificate `{$certBase}` for host `{$this}` expired {$expiration}; web " .


camspiers:
Isn't it a matter of whether strict_types is set in the file that creates the Vhost? If it isn't set in the file that creates the Vhost you can pass in anything you want to name.

bwoebi:
@camspiers uhm , you realize that strict_mode has its effects on call-side not caller-side. The PR is fine.

tpunt:
Not quite. It'll attempt to coerce the value being passed to a string, and according to the ZPP rules, `null` cannot be coerced (https://3v4l.org/nbO3s). Given that `$name` will always be set and can never be `null`, `isset()` will alway be `true`

kelunik:
Code isn't buggy, but bogus. Thanks.

bwoebi:
@kelunik line 69 was buggy though… especially as it never will be null ;-)

kelunik:
Indeed, but `''` should be right now.

camspiers:
I see! You are right :)

@bwoebi I understood that strict_types applies to the caller side, hence why I said "file that creates the Vhost" as in, the file that instantiates it. I was thinking that it would accept null in weak mode, but you are right that it coerce. I haven't used PHP7 really at all (hence my ignorance), I work with Hack almost exclusively.

@bwoebi it's strict_types not strict_mode right?

bwoebi:
@camspiers yeah, strict_types is correct. But anyway… no need to bother with Hack… there's PHP 7 ;-)

camspiers:
Haha. Didn't want to start an argument. Here are a list of things I would really miss if I had to use PHP7: 

* async/await (using generators as amphp does is nice, but having built in support is better).
* type safety
* expressive types (generics, contra and co variance, in general being able to type almost everything)
* XHP with async support
* static type checker (the awesomeness of this can't be overemphasized, I mean really, we have to find out about type bugs at runtime?)
* trait constraints/requirements
* Awesome Collections
* Override and Memoize attributes
* Proper lambdas with a terse syntax
* Constructor argument promotion
* Trailing commas (minor but a nice benefit IMO)
* typed shapes and tuples
* and last and most important of all, I work at Facebook

bwoebi:
Okay, the last point invalidates the whole discussion :-D … Sorry for getting into an argument about Hack with a facebook developer :-D

* my main gripe with async/await is: you loose control. Different Promise/Awaitable interface? No chance. Switch the backend? (Like libevent vs. libuv) not easy either.
* type safety
* expressive types … yeah, let's see what 7.1 brings… Union types are maybe coming… Hope to see some improvements for co/contravariance too. Generics… not sure about these.
* I personally am not a fan of native xml in code. It just feels wrong.
* static type checking though is a feature of the interpreter … an userland project could do it too. I'm excited to maybe once see a static type checker as userland project for php. Unrelated to the Hack language really…
* Your collections are just a poor version of typed arrays IMO. (Well, the Set is nice (because of it's complexity guarantees) … but Vector and Map… mhhh.) … too many specializations too.
* Override attributes don't bring much value I think. Also doesn't help with the mess called late static binding. Memoize sounds nice in theory, though I can't think of many real world cases where it would be helpful.
* Shut up about the short closures/lambdas! RFC which failed was mine… bah! Sometimes it has advantages to not have a formal voting process. _Sometimes_.
* Doesn't ctor argument promotion make it easy to miss the properties of the constructor when the class already has a bunch of other properties?
* I'm undecided on trailing commas. It's rare to have that many arguments that you being properly lining them up. It's rather an indicator for a bad API design then.
* Shapes seem to be more like implicitly typed objects … they're weird; why do you do `shape('x' => 1, 'y' => 1)` instead of like `Point { 'x' => 1, 'y' => 1 }` would be more explicit and it'd even have an explicit type now. No reason to force an array there? And tuples really are just sugar for SPLFixedArrays, no? The only thing nice about them is being able to return a fixed size array with fixed types for each element.
* and last and most important of all, I work on php-src

Okay, yeah, you do have nice things, but sometimes I feel like there's not too much thought behind some features and some features are really just "oh, I think that could be nice, implement it straight ahead". I am amused how you explain syntax choices, but leave the important point, why the feature is needed and the ways in which it is superior to existing syntax.
Sometimes Hack feels like a bit a playground to see live what should _not_ go into PHP :-D

camspiers:
Great points @bwoebi. Thanks for responding. First it's important to point out that I am not a engineer on Hack or HHVM, merely an engineer that uses them a lot by virtue of working at Facebook.

* Good points about async/await, it is a tradeoff. There are benefits to having it implemented as part of the language, but downsides as you mention.
* "I personally am not a fan of native xml in code. It just feels wrong." yeah a lot of people have that reaction to it (in particular JSX in React). But there are a lot of real safety benefits to XHP that I don't think are as readily or easily found in templating/DSL style solutions.
* regarding type checking, yes I agree and it would be wonderful for something like hh_client to be implemented for PHP (but the language is probably a little unsound for it as it stands now, hence why Hack removed a fair bit). But the point remains: it exists for Hack and not for PHP and there are a lot of real world benefits to it.
* I think Set and Pair are most useful, but Collections do enforce some nice constraints that you don't get with arrays.
* Override has a lot of value in a large codebase. It allows you to ensure that an implementation doesn't change "under your feet" 
* Memoize is very useful but must be applied with thought given the serialization and memory usage aspects.
* "Shut up about the short closures/lambdas! RFC which failed was mine" I'm sorry about that :( If I were a voting member I would have voted for it. Though I did find the `~>` syntax an unnecessary deviation from Hack, and wasn't at all convinced by the arguments for it over `==>`
* "Doesn't ctor argument promotion make it easy to miss the properties of the constructor when the class already has a bunch of other properties" I'm not sure what you mean, but possibly. 
* Trailing commas IMO are mostly useful for blame. Allow you to avoid editing lines you don't need to. This helps in a large codebase with many engineers
* Shapes aren't implicitly typed. Shapes are a lot like records in other languages. They are backed by arrays in Hack, but they allow you to provide types for each key. So yes they are like objects in that you can achieve the same thing, but they are backed by arrays. They can also be made opaque which can be useful. Tuples are typed and can be heterogeneous, that they are typed is what most differentiates them from SPLFixedArrays.

>I feel like there's not too much thought behind some features and some features are really just "oh, I think that could be nice, implement it straight ahead"

That definitely isn't the case, most features are implemented for specific reasons derived from experience with an enormous PHP codebase. For example Facebook had a few different eras where async programming was implemented in various ways in PHP, including approaching it in a very similar to Icicle and Amphp using generators for cooperative multitasking. Async/await was introduced to improve and disambiguate the API of the generator approach at the same time as getting a bunch of performance improvements.

PHP was used as a example of what not to allow/implement in Hack so maybe the HHVM/Hack team are just now returning the favor by experimenting with language features. :D

sgolemon:
Last but not least, I work on both php-src and hhvm.

#micdrop

kelunik:
For static type checking, there's [`ircmaxell/tuli`](https://github.com/ircmaxell/Tuli). :-)

krakjoe:
Unless you work at facebook, or a small number of other companies, hack programmers are a kind of new species ... Like a rare whale, we dunno much about them ...

It was interesting to hear what would be missed, from someone who works with hack all the time. Some of the things on the list are appealing, maybe ...

camspiers:
Just to be clear, I am not ragging on PHP. I love PHP and am very excited for PHP7. I worked with PHP most days for around 9 years before using Hack, so PHP has a special place in my heart :D. I'm also having a lot of fun using Aerys at the moment, it is a great project.

sgolemon:
Dude, @krakjoe, did you just call me a whale? ;p

kelunik:
@camspiers Do you have anything we can already look at?

@sgolemon :-D

camspiers:
@kelunik Nothing worth looking at. Building a game.

staabm:
feels like this topic misses some meme

krakjoe:
@sgolemon little bit ... not on purpose, but that's what happened ...

sgolemon:
Biggest mo-fo on the block? I can deal with that.

PR #59, opened 2015-11-06
Fixes problem with 'Undefined property' in VhostContainer

dol:
Minor variable case mixing causing PHP 7 to trigger an error: Undefined property


kelunik:
:+1:

PR #60, opened 2015-11-06
Allow async I/O in config files

kelunik:
This allows fetching e.g. configuration and building hosts based on it.

### DIFF
--- a/lib/Bootstrapper.php
+++ b/lib/Bootstrapper.php
@@ -16,16 +16,24 @@ class Bootstrapper {
      *
      * @param \Aerys\Logger $logger
      * @param \Aerys\Console $console
-     * @return \Aerys\Server
+     * @return \Generator
      */
-    public function boot(Logger $logger, Console $console): Server {
+    public function boot(Logger $logger, Console $console): \Generator {
         $configFile = $this->selectConfigFile((string) $console->getArg("config"));
         $logger->info("Using config file found at $configFile");
-        if (!include($configFile)) {
+
+        $returnValue = include $configFile;
+
+        if (!$returnValue) {
             throw new \DomainException(
                 "Config file inclusion failure: {$configFile}"
             );
-        } elseif (!defined("AERYS_OPTIONS")) {
+        } elseif ($returnValue instanceof Promise) {
+            // allow async I/O in config files and wait for it
+            yield $returnValue;
+        }
+
+        if (!defined("AERYS_OPTIONS")) {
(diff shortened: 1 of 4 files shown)


kelunik:
@rdlowrey Maybe we should adjust the failure rate on Coveralls to `-0.25%`.

benjamingr:
Or maybe write a test for it :D?

kelunik:
Yes. :-D

kelunik:
Maybe I should just enable the "require all checks passed before merge" option, leads to better PRs. :-)

lib/Bootstrapper.php
```diff
@@ -16,16 +16,24 @@ public function __construct(callable $hostAggregator = null) {
...
      */
-    public function boot(Logger $logger, Console $console): Server {
+    public function boot(Logger $logger, Console $console): \Generator {
         $configFile = $this->selectConfigFile((string) $console->getArg("config"));
         $logger->info("Using config file found at $configFile");
-        if (!include($configFile)) {
+
+        $returnValue = include $configFile;
+
+        if (!$returnValue) {
```
staabm: wouldnt it lead to a problem when the config doesnt return anything?

kelunik: In that case `(int) 1` is returned for success and `false` for failure.

test/BootstrapperTest.php
```diff
@@ -0,0 +1,30 @@
...
+use function Amp\resolve;
+use function Amp\wait;
+
+class BootstrapperTest extends \PHPUnit_Framework_TestCase {
+    /**
+     * @test
+     * @expectedException \DomainException
+     * @expectedExceptionMessage No config file found, specify one via the -c switch on command line
+     */
+    public function throwsWithoutConfig() {
```
bwoebi: This test has a problem: it may spuriously fail in case you have a config in the default locations.

kelunik: I know, but I'd say we should disable default locations anyway.

bwoebi:
In general, PHP doesn't allow you to yield inside a file. So, when we want async I/O in config file, we need to do a big `return (function() { /* .. */ })();`, right?

test/BootstrapperTest.php
bwoebi: I don't disagree… let me do this and merge this PR.

kelunik:
You missed the coroutine wrapper, but right.

Issue #27, opened 2015-01-30
Support application hot-reload


rdlowrey:
Listen over local TCP loopback for hot reload commands (enable with a port number option switch passed to bin/aerys) or unix domain socket where available.

kelunik:
We need also support for internal reload, so `Server::reload`.

morrisonlevi:
Do we already have a protocol format for sending non-http messages to the server? I'm not aware of one, but this project has been moving really fast lately :D

bwoebi:
No, we don't. We probably should end up doing something like `bin/aerys -c config.php --restart`, save an unix domain socket (or file with a random port number from the ephermal port range) to `"/tmp/".str_replace("/", "_", $config_path)` (we should be additionally able to guarantee that only one aerys instance of the same config runs now). Then we'll be able to use a very simple length-prefixed text protocol.

bwoebi:
Fixed by #62.

kelunik:
Where's `Server::reload`? ;-P

bwoebi:
Use the `CommandClient` to stop all, to just restart the specific instance we already have `Server::doStop()`.

PR #62, opened 2015-11-08
Add graceful restarting

bwoebi:
Fixing #27

Review requested… \cc @kelunik

### DIFF
--- a/lib/CommandClient.php
+++ b/lib/CommandClient.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace Aerys;
+
+class CommandClient {
+    private $buf;
+    private $promisors = [];
+    private $path;
+    private $sock;
+    private $writer;
+    private $written = 0;
+    private $writeWatcher;
+
+    public function __construct($config) {
+        $this->path = sys_get_temp_dir()."/aerys_".str_replace(["/", ":"], "_", Bootstrapper::selectConfigFile($config)).".tmp";
+        $this->writer = (new \ReflectionClass($this))->getMethod("writer")->getClosure($this);
+    }
+
+    private function send($msg) {
+        if (!$this->sock) {
+            $this->establish();
+        } elseif ($this->writeWatcher) {
+            \Amp\enable($this->writeWatcher);
+        }
+        $msg = json_encode($msg);
+        $this->buf .= pack("N", \strlen($msg)) . $msg;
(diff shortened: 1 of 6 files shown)


lib/DebugProcess.php
```diff
@@ -14,6 +14,11 @@ public function __construct(Logger $logger, Bootstrapper $bootstrapper = null) {
     }
 
     protected function doStart(Console $console): \Generator {
+        if ($console->isArgDefined("restart")) {
+            $this->logger->critical("You cannot restart a debug aerys instance via command");
```
staabm: Why?

lib/CommandClient.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->sock = $this->writeWatcher = null;
+
+        $promisors = $this->promisors;
+        $this->promisors = [];
+        foreach ($promisors as $deferred) {
+            $deferred->fail(new \Exception("Couldn't write command, server failed."));
+        }
+    }
+
+    public function restart() {
```
staabm: Return type hint

lib/WatcherProcess.php
```diff
@@ -241,7 +347,7 @@ private function parser($ipcClient): \Generator {
     private function onDeadIpcClient(string $readWatcherId, $ipcClient) {
         \Amp\cancel($readWatcherId);
         @fclose($ipcClient);
-        unset($this->ipcClients[(int)$ipcClient]);
+        unset($this->ipcClients[$readWatcherId]);
```
staabm: Unrelated bugfix?

kelunik:
This needs at least one new test. ;-P

lib/DebugProcess.php
bwoebi: Because we don't have separate processes for debug instances. Restarting is actually killing workers and starting them again.

staabm: Sure it is but does it matter for the user? It only matters that the sources get reloaded..?

bwoebi: The point is that you can't restart it. Hence the message. That's all.

lib/WatcherProcess.php
bwoebi: Not a bugfix, just a cosmetic change.

bwoebi:
Merged via 12bdafe.

PR #71, opened 2016-01-20
Deprecate AERYS_OPTIONS

bwoebi:
What do you think about not using a constant?

It has the advantage of not having to define() the array (gives you more freedom when assembling options) … but maybe it is a fine as is as it forces the options to be centralized more or less in one single place.

So, please think about it … I'm unsure too.


kelunik:
It's a design decision to have it as a constant so everything must be at one place.

staabm:
Actually I like the approach we had before this PR. I think we should only change it after we discover problems which cannot be solved with the current mechanism or requires workarounds

bwoebi:
Well, then … and as I'm anyway unsure and just wanted some input… REJECTED! :-)

PR #81, opened 2016-02-16
added phpdocs to InternalRequest

staabm:
to support IDE autocompletion

### DIFF
--- a/lib/InternalRequest.php
+++ b/lib/InternalRequest.php
@@ -7,28 +7,52 @@ use Amp\Struct;
 class InternalRequest {
     use Struct;
 
+    /** @var Client */
     public $client;
+    /** @var Vhost */
     public $vhost;
+    /** @var callable|null */
     public $preAppResponder;
+    /** @var Response */
     public $response;
+    /** @var \Generator */
     public $responseWriter;
+    /** @var array */
     public $badFilterKeys = [];
+    /** @var boolean */
     public $filterErrorFlag;
+    /** @var boolean */
     public $isServerStopping;
+    /** @var integer */
     public $streamId;
+    /** @var string|array */
     public $trace;
+    /** @var string */
     public $protocol;
+    /** @var string */
     public $method;
+    /** @var array */
     public $headers;
+    /** @var Body */
     public $body;
+    /** @var string */
     public $uri;
+    /** @var string */
     public $uriRaw;
+    /** @var string */
     public $uriHost;
+    /** @var integer */
     public $uriPort;
+    /** @var string */
     public $uriPath;
+    /** @var string */
     public $uriQuery;
+    /** @var array */
     public $cookies;
+    /** @var int */
     public $time;
+    /** @var string */
     public $httpDate;
+    /** @var array */
     public $locals;
 }


lib/InternalRequest.php
```diff
@@ -7,28 +7,51 @@
...
     public $responseWriter;
+    /** @var array */
     public $badFilterKeys = [];
+    /** @var boolean */
     public $filterErrorFlag;
+    /** @var boolean */
     public $isServerStopping;
+    /** @var string */
     public $streamId;
     public $trace;
```
staabm: I couldnt figure out what the correct type should be for `$trace`

bwoebi: string.

staabm: fixed

lib/InternalRequest.php
```diff
@@ -7,28 +7,52 @@
 class InternalRequest {
     use Struct;
 
+    /** @var Client */
     public $client;
+    /** @var Vhost */
     public $vhost;
+    /** @var callable */
```
staabm: is `callable` fine?

bwoebi: yes, callable or null

staabm:
here we go

lib/InternalRequest.php
```diff
@@ -7,28 +7,52 @@
...
+    /** @var \Generator */
     public $responseWriter;
+    /** @var array */
     public $badFilterKeys = [];
+    /** @var boolean */
     public $filterErrorFlag;
+    /** @var boolean */
     public $isServerStopping;
+    /** @var string */
     public $streamId;
```
bwoebi: streamId is integer

lib/InternalRequest.php
```diff
@@ -7,28 +7,52 @@
...
+    /** @var Body */
     public $body;
+    /** @var string */
     public $uri;
+    /** @var string */
     public $uriRaw;
+    /** @var string */
     public $uriHost;
+    /** @var string */
     public $uriPort;
```
bwoebi: integer

lib/InternalRequest.php
staabm: I somehow had in mind that we discussed that this needs to be a string so it can "never" overflow?

bwoebi: no, that's not stream ids, that's about ping counter.

kelunik: They're per connection? When do I use these IDs?

bwoebi: you don't, it's an internal detail to HTTP/2

staabm: changed

lib/InternalRequest.php
```diff
@@ -7,28 +7,52 @@
 class InternalRequest {
     use Struct;
 
+    /** @var Client */
     public $client;
```
staabm: @bwoebi should those fields be properly typed instead of using phpdoc?

bwoebi: what do you mean with properly typed here?

staabm: use `public Client $client` instead of a phpdoc (where appropriate in this class)

kelunik: We don't have typed properties yet.

staabm: its a pity that I spent so less time with php7 that I even dont know what the language actually supports. :-/

PR #82, opened 2016-02-16
Use always strings to represent the port

staabm:
before it a string when result of parse_url() and a int when the
default was used.


bwoebi:
I recommend it the other way round …force always integer.

staabm:
@bwoebi adjusted per feedback

PR #83, opened 2016-02-22
Use unix domain sockets when available, resolves #28

### LINKED ISSUE #28: Use unix domain sockets for IPC when available
Currently all inter-process communication between the main server process and workers is conducted over TCP on the local loopback address. Use unix domain sockets when possible.

kelunik:


bwoebi:
Rebased into https://github.com/amphp/aerys/commit/7a5f42c7149a85596b7f3ee65ae0fb338ad6936f

PR #87, opened 2016-04-16
Fix demo.php: Replace 'send' method calls with 'end' method calls.

rlmumford:
### Motivation

I installed aerys using this guide http://blog.kelunik.com/2015/10/21/getting-started-with-aerys.html, but when running the demo application I got fatal errors because there is no longer a send() method on the StandardResponse class.

### Solution

I have replaced the calls to send() with calls to end().

### DIFF
--- a/demo.php
+++ b/demo.php
@@ -17,14 +17,14 @@ const AERYS_OPTIONS = [
 
 $router = router()
     ->get("/", function(Request $req, Response $res) {
-        $res->send("<html><body><h1>Hello, world.</h1></body></html>");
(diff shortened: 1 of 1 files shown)


bwoebi:
Thanks for that, totally missed that one when removing it :-)

kelunik:
Why does Coveralls use comments and the status API. -.-

kelunik:
@rlmumford Thanks for the heads up, had it on my list, but didn't actually update the posts yet. Did it now  with https://github.com/kelunik/kelunik.github.io/commit/e2102297b8278f1f6779be22c6ec92b13f779286.

PR #94, opened 2016-06-03
Handle MIME types with suffixes correctly

joshdifabio:
`Aerys\Root` sets the MIME type for HTTP responses by looking up the file extension of the requested file in the [mime file](https://github.com/amphp/aerys/blob/v0.4.3/etc/mime). However, when `Aerys\Root` parses the mime file, the regex used ignores MIME type suffixes. This means that for `.svg` files, Aerys response with a content type of `image/svg` instead of `image/svg+xml`, which prevents browsers from displaying `svg` files correctly. This PR addresses this issue.

### DIFF
--- a/lib/Root.php
+++ b/lib/Root.php
@@ -665,7 +665,7 @@ class Root implements ServerObserver {
                 "Failed loading mime associations from file {$mimeFile}"
             );
         }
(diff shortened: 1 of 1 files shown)


kelunik:
Could you add a test for it?

bwoebi:
@joshdifabio May you provide the test please?

Issue #96, opened 2016-07-01
Server has type declaration for Aerys\Logger instead of Psr\LoggerInterface


cspray:
I would like to be able to pass in my own implementation of a logger to the Aerys Server. This would allow me to persist logs to file, in a non-blocking manner of course, or send logs for certain levels to error monitoring services, for example. Is there an internal implementation reason why this type declaration is for a concrete class and not the interface? If so I could be able to provide a PR with the appropriate work if the problem can be described.

By the way, I think Aerys is pretty damn awesome. Writing an async app being able to yield Promises was an enlightening moment and really helped me better understand generators. It makes me cringe now when I have to work in JS and the callbacks that you all-too-often have to deal with.

kelunik:
I don't see any reason for this. Could you provide a PR to change the type declarations?

bwoebi:
Thanks, merged in #97.

PR #97, opened 2016-07-01
Server now declares Psr\Log\LoggerInterface instead of concrete class

cspray:
See https://github.com/amphp/aerys/issues/96

### DIFF
--- a/test/WebsocketParserTest.php
+++ b/test/WebsocketParserTest.php
@@ -16,6 +16,7 @@ use Aerys\VhostContainer;
 use Aerys\Websocket;
 use Aerys\Websocket\Code;
 use Aerys\Websocket\Rfc6455Endpoint;
+use Psr\Log\LoggerInterface as PsrLogger;
 
 class WebsocketParserTest extends \PHPUnit_Framework_TestCase {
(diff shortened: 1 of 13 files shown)


kelunik:
There are a few more occurrences: https://github.com/amphp/aerys/search?utf8=%E2%9C%93&q=Logger

Could you fix them as well?

cspray:
@kelunik oops, sorry about that. Not sure what I was thinking. Clearly I shouldn't code before I drink coffee. Will hit those in a few.

test/WebsocketParserTest.php
```diff
@@ -16,6 +16,7 @@
 use Aerys\Websocket;
 use Aerys\Websocket\Code;
 use Aerys\Websocket\Rfc6455Endpoint;
+Use Psr\Log\LoggerInterface as PsrLogger;
```
cspray: Inappropriate capital 'U'

kelunik: Hm?

staabm: First char in the line

kelunik: Oh, missed that.

PR #100, opened 2016-07-28
Include full DateTime for Logger

cspray:
Pretty self explanatory. Would like to get the full datetime for log entries.


bwoebi:
No reason why not… merging :-)

PR #104, opened 2016-10-20
added missing argument

strayobject:
as per method signature:
`private function tryApplication(InternalRequest $ireq, callable $application, array $filters)`


bwoebi:
Changing method signatures and forgetting to change them all… ;-)

Thanks for the PR :+1:

Issue #105, opened 2016-10-20
onClose event not fired or not handled


lantonello:
I'm trying the Aerys and... wow, this is great! Thanks for that!
But there is a problem with onClose event.

I have installed with composer required and get the kelunik/demo-chat for testing. This is my config file:

```
<?php
use Aerys\Host;
use Aerys\Router;
use Yuudy\Libs\Chat;
use function Aerys\root;
use function Aerys\websocket;

/* --- Global server options -------------------------------------------------------------------- */
const AERYS_OPTIONS = [
    "keepAliveTimeout" => 60,
];

/* --- http://localhost:8822/ ------------------------------------------------------------------- */
$router = (new Router())->route('GET', 'ws', websocket(new Chat));
(new Host())->name('localhost')->expose('*', 8822)->use($router);
```

My Yuudy\Libs\Chat class is a copy of Kelunik\Demo\Chat class. I only change the namespace. Everthing is ok. The Aerys terminal window (with -d flag) show that the server is online and log all connections.

On javascript side, assuming a var conn is a WebSocket object, nothing happens when I call:

`conn.close();`

In Firefox console, after 10 or 12 seconds, I see the "error" and "close" events of javascript implementation. In Aerys terminal window nothing happen. I think the onClose server event are not fired or not handled.

PS. Sorry for my bad english.

kelunik:
@lantonello Do you mean that no `close` event is being logged? That might be true. The `onClose` in `Websocket` instances should still be executed.

lantonello:
@kelunik Yes, no `close` event is being logged. Sorry for typo.
The `onClose` method in `Websocket` instance are not executed when I call `conn.close()` with javascript.
When I made this call, an error ocurred in Firefox console: The connection has interrupted (in red)
After 10 or 12 seconds, I see the `error` and close `events` in Firefox console.

kelunik:
`onClose` will be executed, it will just take its time. @bwoebi I think we should improve that, it takes really long if we always wait for the timeout.

It doesn't log a `close` event, because it logs an `export` event. That means that the server gives control of the socket to somebody else, that's `Rfc6455Endpoint` here. And `Rfc6455Endpoint` doesn't log anything yet.

bwoebi:
Seems like there was another issue… here's the real fix for the particular issue you've probably encountered...

See https://github.com/amphp/aerys/commit/f9724bb2f52f16102bb542de798a44a9976cdb23

PR #106, opened 2016-10-21
Log socket close event for exported sockets

kelunik:
Relates to #105.


bwoebi:
Hmm, that's preserving $client and all the properties it contains. I'll push a better fix tomorrow.

kelunik:
@bwoebi I adjusted the patch to `use` only `clientAddr` and `clientPort`.

PR #111, opened 2016-12-05
Simplify CPU counter to use nproc

kelunik:


bwoebi:
Is this available on all distros?

kelunik:
I'm not sure on that, but as it's on coreutils as you say, it should be fine.

kelunik:
A advantage of `nproc` is that it shows the count of processors available to the process instead of the total installed number.

bwoebi:
Well, for example it is not included with busybox … Well, I'm not sure, I let you decide whether to merge.

lt:
While detecting available processors rather than installed processors seems to make perfect sense, I'm not sure I've ever seen a scenario where processors have been deliberately taken offline (which is what I believe is being catered for with nproc, rather than some kind of software limitation on processor usage)

staabm:
any hyperthreading related things we need to consider?

morrisonlevi:
All processors are routinely not available inside of cgroups and cpusets. I believe in the former case that `nproc` it works but I have no idea about the second.

lt:
```
$ nproc
8
$ docker exec -ti test nproc
7
```
@morrisonlevi is right :)

bwoebi:
Fine, merging that for now, when it explodes somewhere, we can still revert it.

kelunik:
@morrisonlevi @lt Is it a good thing to rely on `nproc` then? I guess so?

Issue #113, opened 2016-12-16
Allow Router to accept multiple HTTP methods at once for a route


broiniac:
Let's say I want to build subpage with form. Let's call it `/product/create`. I also want to handle both subpage display as well as form handling in one callable. Here is my code for routing definition:
```
        $router = router();
        $router->route(
            'GET', //method
            '/product/create', //url
            'App\\Controllers\\ProductController::createAction' //callable
        );
        $router->route(
            'POST', //method
            '/product/create', //url
            'App\\Controllers\\ProductController::createAction' //callable
        );
```
I think it's best way right now (besides using $router->post() / $router->get() methods);

$router->route() method itself as first parameter takes string. I believe router will be more coder-friendly, if it will be allowed to send array of methods as first param as well as string.

Example below:
```
        $router = router();
        $router->route(
            ['GET', 'POST'], //CHANGE HERE: methods array
            '/product/create', //url
            'App\\Controllers\\ProductController::createAction' //callable
        );
        $router->route(
            'GET', //REGULAR WAY SHOULD WORK AS WELL
            '/product/show, //url
            'App\\Controllers\\ProductController::showAction' //callable
        );
```

I will create pull request with proposed changes in few minutes.

kelunik:
Having one handler doing multiple things is against the single responsibility principle. I'm against making it easier to do the wrong thing.

bwoebi:
@kelunik It depends on what you want to do there, whether it violates SRP or not.

It's quite a common request, and I feel like we should honor it.

kelunik:
@broiniac Why do you have multiple methods registered to one handler? What's the use case?

kelunik:
@bwoebi There are very few things that don't violate the SRP, one example is making an alias like `POST` → `PATCH` like Twitter does, but such things are better handled by just defining a middleware that changes the request method.

broiniac:
@kelunik As I said before, example use case might be situation, when I want to handle both display and form handling in one callable hidden under same url.

If it's violating SRP in that way, there always can be some separate method for this one, like "routeWithManyMethods" or simple "routes" (name of the latter can be confusing though).

[...]

PR #115, opened 2016-12-16
#113: allow to set multiple methods for single route at once

broiniac:


lib/Router.php
```diff
@@ -242,21 +242,21 @@ public function __call(string $method, array $args): Router {
      * the trailing slash. Temporary redirects are used to redirect to the canonical URI
      * (with a trailing slash) to avoid search engine duplicate content penalties.
      *
-     * @param string $method The HTTP method verb for which this route applies
+     * @param array $method The HTTP method verb for which this route applies
```
bwoebi: This is `string|array` now, not `array`, right?

lib/Router.php
```diff
@@ -266,19 +266,19 @@ public function route(string $method, string $uri, ...$actions): Router {
         }
 
         $actions = array_merge($this->actions, $actions);
+        $methods = is_array($methods) ? $methods : [$methods];
```
bwoebi: Just cast to array here, i.e. in the foreach: `foreach ((array) $methods as $method)`

kelunik:
The change does no longer check that methods are strings now.

lib/Router.php
broiniac: Ok

kelunik: The current approach is actually the cleaner one. `(array) $methods` will also happily cast objects to an array instead of putting the object into that array.

broiniac:
@kelunik I was hoping it is done in line 257:
```
        if ($methods === "" || $methods === []) {
```

lib/Router.php
```diff
@@ -242,13 +242,13 @@ public function __call(string $method, array $args): Router {
...
      *
-     * @param array $method The HTTP method verb for which this route applies
+     * @param string|array $methods The HTTP methods verb or array of verbs for which this route applies
      * @param string $uri The string URI
      * @param Bootable|Middleware|Monitor|callable ...$actions The action(s) to invoke upon matching this route
      * @throws \DomainException on invalid empty parameters
      * @return self
      */
-    public function route($methods, string $uri, ...$actions): Router {
+    public function route(mixed $methods, string $uri, ...$actions): Router {
```
kelunik: Uhm, `mixed` does not exist.

lib/Router.php
```diff
@@ -292,12 +291,12 @@ public function route($methods, string $uri, ...$actions): Router {
                 $response->setHeader("Content-Type", "text/plain; charset=utf-8");
                 $response->end("Canonical resource URI: {$path}");
             };
-            foreach ($methods as $method) {
+            foreach ((array) $methods as $method) {
```
kelunik: Please add a check that `$method` is a string.

kelunik:
@broiniac How does that check that the items in the array are strings?

lib/Router.php
broiniac: Casting to string will be ok, or should I use condition?

kelunik: I'd use a check. Something other than a string there is an error.

rdlowrey:
I really don't like allowing multiple methods for a single route like this. It's so trivial to do this in userland and it's not generally a good idea from an HTTP perspective to be using the same code for multiple methods.

bwoebi:
Closed as per #113

broiniac:
OK than, no hard feelings :)

Issue #112, opened 2016-12-15
Upgrade deflateResponseFilter() to support non-text/* MIME types


joshdifabio:
It'd be great to be able to deflate _application/x-javascript_, _application/json_, etc. Javascript files in particular are often very weighty, even when minified.

Currently [there's a todo for this within `deflateResponseFilter()`](https://github.com/amphp/aerys/blob/9cae6c9fd93136b1db33f770e0a981573b63c220/lib/functions.php#L424); I'm just expressing interest in seeing it implemented.

kelunik:
Question is which ones should be compressed.

joshdifabio:
> Question is which ones should be compressed.

The todo in the code seems to propose a new configuration option to allow the end user to define that, which would be consistent with nginx and Apache.

bwoebi:
How to best implement that? An option which accepts a list with allowed prefixes?

kelunik:
Content types I guess, not prefixes

bwoebi:
@kelunik I mean by prefix of the content type. I.e. If you want to straight ahead deflate everything with a text/* mime, you just pass in "text/" instead of tens of combinations.

kelunik:
Yes, and I think we should not do that. We should compress `text/*` by default and then additional content types present in the config. I don't think prefixes make sense aside from `text/`.

staabm:
You will want to compress text based format, like css/js/svg but not files etc... I guess some sane default can match 90% of the use-cases

kelunik:
@staabm The issue is that JS is usually served as `application/x-javascript` instead of `text/javascript`.

I think we should always compress `text/*` and default the option to `application/x-javascript` and `application/json`.

staabm:
with text-based I did not technically meant text/* but "everything which is regular text content".

I guess something like
- text/*
- */xml
- /*+xml
- /*javascript

could do as a default

bwoebi:
Thus I think a simple regex should express that easily?

(i.e. `#^(?:text/.*+|[^/]*+/xml|[^+]*\+xml|application/(?:json|(?:x-)?javascript))$#`)

Then you can customize as much as you want to…

joshdifabio:
I think regex support would be good.

kelunik:
[...]

PR #117, opened 2016-12-16
Add option to match content-type against a regex for deflateResponseFilter()

bwoebi:
See #112

### DIFF
--- a/lib/Options.php
+++ b/lib/Options.php
@@ -7,6 +7,8 @@ use Amp\Struct;
 class Options {
     use Struct;
 
+    const MAX_DEFLATE_ENABLE_CACHE_SIZE = 1024;
+
     private $debug;
     private $user = null;
     private $maxConnections = 1000;
@@ -23,6 +25,7 @@ class Options {
     private $maxFramesPerSecond = 60;
     private $allowedMethods = ["GET", "POST", "PUT", "PATCH", "HEAD", "OPTIONS", "DELETE"];
     private $deflateEnable;
+    private $deflateContentTypes = '#^(?:text/.*+|[^/]*+/xml|[^+]*\+xml|application/(?:json|(?:x-)?javascript))$#i';
     private $configPath = null;
     
     private $maxFieldLen = 16384; // this must be strictly less than maxBodySize
@@ -40,10 +43,16 @@ class Options {
 
     private $shutdownTimeout = 3000; // milliseconds
 
+    private $_dynamicCache;
+
     private $_initialized = false;
 
     public function __construct() {
         $this->deflateEnable = \extension_loaded("zlib");
+
+        $this->_dynamicCache = new class {
+            public $deflateContentTypes = [];
+        };
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lib/Options.php
```diff
@@ -40,10 +43,16 @@ class Options {
 
     private $shutdownTimeout = 3000; // milliseconds
 
+    private $dynamic_cache;
```
kelunik: Should be `$dynamicCache`.

kelunik: Why a dynamic cache in Options? Options should be a pure value object.

test/deflateResponseFilterTest.php
```diff
@@ -13,6 +13,7 @@ function testNoDeflate() {
         $ireq = new InternalRequest;
         $ireq->client = new Client;
         $ireq->client->options = new Options;
+        $ireq->client->options->deflateContentTypes = "#^text/.*$#";
```
kelunik: Why do we need that?

bwoebi: This tests setting the option

lib/Options.php
bwoebi: It's intentionally an underscore to distinguish it from the actual options.

Where to put it instead so that it is accessible from deflateResponseFilter()?
And yeah, Options is a value object, but value objects may contain dynamic entities…

lib/functions.php
```diff
@@ -420,14 +420,26 @@ function deflateResponseFilter(InternalRequest $ireq): \Generator {
...
+    if (!$doDeflate = $options->dynamic_cache->deflateContentTypes[$headers["content-type"][0]] ?? null) {
+        if ($doDeflate === 0) {
+            return $headers;
+        }
+
+        if (count($options->dynamic_cache->deflateContentTypes) == Options::MAX_DEFLATE_ENABLE_CACHE_SIZE) {
+            unset($options->dynamic_cache->deflateContentTypes[key($options->dynamic_cache->deflateContentTypes)]);
+        }
+
+        $doDeflate = preg_match($options->deflateContentTypes, $headers["content-type"][0]);
```
kelunik: This one needs changes. You have to split the `content-type` header, as it may contain additional items, such as the charset.

lib/Options.php
bwoebi: yeah, prefixing would've been better, I agree.

lib/functions.php
bwoebi: Good point. Fixed.

lib/Options.php
```diff
@@ -40,10 +43,16 @@ class Options {
...
+    private $_dynamicCache;
+
     private $_initialized = false;
 
     public function __construct() {
         $this->deflateEnable = \extension_loaded("zlib");
+
+        $this->_dynamicCache = new class {
+            public $deflateContentTypes = [];
+        };
```
kelunik: Do we have any reason for this additional encapsulation?

bwoebi: Yes. The Options class itself is immutable.

kelunik: It's not by definition as long as it has mutable contents.

bwoebi: Uhm, I mean that __set() is preventing direct writes after initialization.
And I didn't feel like adding an explicit exception for _dynamicCache right there.

bwoebi:
Merged as 544ee5f152db53f8a0fc79644131d13ac68d36a3

PR #127, opened 2017-01-24
Split send method

trowski:
This PR splits `Endpoint::send()` into three methods:
 - `Endpoint::send(string $data, int $clientId)` for sending messages to a single client.
 - `Endpoint::broadcast(string $data, array $exceptIds = null)` for sending to all clients (optionally excluding those given in the array).
- `Endpoint::multicast(string $data, array $clientIds)` for sending to a set of specific clients.

`Endpoint::sendBinary()` is similarly split into three methods, adding `Endpoint::broadcastBinary()` and `Endpoint::multicastBinary()`.

Having three distinct methods for sending messages to a single client, all clients, or to a set of clients makes it easier to determine where messages are going at a glance.

### DIFF
--- a/lib/Websocket/Endpoint.php
+++ b/lib/Websocket/Endpoint.php
@@ -33,7 +33,7 @@ interface Endpoint {
      *
      * @return \AsyncInterop\Promise<int>
      */
-    public function simulcast(string $data, array $clientIds): Promise;
+    public function multicast(string $data, array $clientIds): Promise;
 
     /**
      * Send a binary message to the given client(s).
@@ -63,7 +63,7 @@ interface Endpoint {
      *
      * @return \AsyncInterop\Promise<int>
      */
-    public function simulcastBinary(string $data, array $clientIds): Promise;
+    public function multicastBinary(string $data, array $clientIds): Promise;
 
     /**
      * Close the client connection with a code and UTF-8 string reason.
--- a/lib/Websocket/Endpoint.php
+++ b/lib/Websocket/Endpoint.php
@@ -6,24 +6,44 @@ use AsyncInterop\Promise;
 
 interface Endpoint {
     /**
-     * Send a UTF-8 text message to the given client(s).
+     * Send a UTF-8 text message to the given client.
      *
-     * @param int|array|null $clientId Single client ID to send data to, an array of client IDs, or null for all clients.
      * @param string $data Data to send.
+     * @param int $clientId
      *
      * @return \AsyncInterop\Promise<int>
      */
-    public function send(/* int|null|array */ $clientId, string $data): Promise;
+    public function send(string $data, int $clientId): Promise;
+
+    /**
+     * Broadcast a UTF-8 text message to a set of clients.
+     *
+     * @param string $data Data to send.
+     * @param int[]|null $clientIds Array of client IDs or null for all clients.
+     *
+     * @return \AsyncInterop\Promise<int>
+     */
+    public function broadcast(string $data, array $clientIds = null): Promise;
 
     /**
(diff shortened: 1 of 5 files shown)


lib/Websocket/Endpoint.php
```diff
@@ -6,24 +6,64 @@
...
+    /**
+     * Send a UTF-8 text message to a set of clients.
+     *
+     * @param string $data Data to send.
+     * @param int[]|null $clientIds Array of client IDs.
      *
      * @return \AsyncInterop\Promise<int>
      */
-    public function send(/* int|null|array */ $clientId, string $data): Promise;
+    public function simulcast(string $data, array $clientIds): Promise;
```
staabm: Shouldnt it be `multicast` ?

trowski: Yeah, `multicast` is better. Naming is hard. :smile:

kelunik:
Huge :+1:, should we have separate binary methods or `Endpoint::TYPE_BINARY` and `Endpoint::TYPE_TEXT`?

lib/Websocket/Rfc6455Gateway.php
```diff
@@ -573,39 +573,58 @@ private function write(Rfc6455Client $client, string $msg, int $opcode, int $rsv
...
         }
+        return $this->compile($client, $data, $opcode);
+    }
 
-        if (\is_array($clientId)) {
+    public function broadcast(/* ?array */ $exceptIds, string $data, bool $binary): Promise {
+        if (empty($exceptIds)) {
             $promises = [];
-            foreach ($clientId as $id) {
+            foreach (array_keys($this->clients) as $id) {
```
kelunik: Isn't `foreach ($this->clients as $id => $client)` better here, as that doesn't copy the array?

test/WebsocketTest.php
```diff
@@ -388,7 +388,7 @@ function onData(int $clientId, Websocket\Message $msg) {
     function testFragmentation() {
         Loop::execute(\Amp\wrap(function () {
             list($endpoint, $client, $sock, $server) = yield from $this->initEndpoint(new NullWebsocket);
-            $endpoint->send(null, str_repeat("*", 131046), true)->when(function() use ($sock, $server) { stream_socket_shutdown($sock, STREAM_SHUT_WR); $server->requireClientFree = true; });
+            $endpoint->broadcast(null, str_repeat("*", 131046), true)->when(function() use ($sock, $server) { stream_socket_shutdown($sock, STREAM_SHUT_WR); $server->requireClientFree = true; });
```
kelunik: There shouldn't be a `null` here I guess.

rdlowrey:
Can't review til later but 👍

lib/Websocket/Rfc6455Gateway.php
trowski: Definitely, leftover from the old way `send` was implemented.

test/WebsocketTest.php
trowski: Here `$endpoint` is actually an instance of `Rfc6455Gateway`. I didn't change the argument order of that internal class or the naming in the tests. I'll update both.

bwoebi:
I like this.
Initially I thought less API would be better, but here the balance was shifted too far away from easily understandable API.

Please adopt Niklas' suggestions, then I'll merge this.

bwoebi:
@kelunik I prefer separate binary methods. That's a cleaner approach than flags IMHO. Also better for extensibility without having to resort to multiple optional parameters.

trowski:
@bwoebi Updated.

@kelunik I agree with @bwoebi in preferring separate methods vs. flags.

PR #130, opened 2017-02-06
Add default reason phrase if none is set

### LINKED ISSUE #129: WebSocket server doesn't give "Switching protocols" reason phrase during handshake
When responding to a HTTP upgrade request from a new WebSocket client, the WebSocket server doesn't give the human-readable text (or "reason") in the HTTP status line.

The server's response looks like:

```
HTTP/1.1 101
upgrade: websocket
connection: upgrade
sec-websocket-accept: ...
```

I understand it's debatable as to whether [RFC6455 Section 4.2.2](https://tools.ietf.org/html/rfc6455#section-4.2.2) and [RFC2616 Section 6.1](https://tools.ietf.org/html/rfc2616#section-6.1.1) define the "reason" or "reason phrase" as mandatory.

However, the lack of "reason" breaks some WebSockets client libraries. In the interests of supporting other implementations, it'd be great if we could send this text.

The conventional "reason" for a 101 status code is "Switching protocols", already defined in [lib/constants.php](https://github.com/amphp/aerys/blob/master/lib/constants.php#L10). It looks like commit 0b34e29f already addressed adding a default "reason" text for standard HTTP response codes (ie issue #25), but somehow this isn't used in the WebSocket handshake response.

kelunik:
Resolves #129.

### DIFF
--- a/lib/StandardResponse.php
+++ b/lib/StandardResponse.php
@@ -8,7 +8,7 @@ class StandardResponse implements Response {
     private $state = self::NONE;
     private $headers = [
         ":status" => 200,
-        ":reason" => null
+        ":reason" => null,
     ];
     private $cookies = [];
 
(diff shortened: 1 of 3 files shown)


kelunik:
@trowski Can you merge this?

PR #136, opened 2017-02-27
Move fastroute to suggestion

tyx:
As it is not mandatory to use aerys

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,14 +36,17 @@
         "amphp/file": "^0.1.1",
         "amphp/socket": "^0.9",
         "league/climate": "^3",
-        "nikic/fast-route": "^1",
         "psr/log": "^1"
+    },
+    "suggest": {
+        "nikic/fast-route": "To use embedded router"
     },
     "require-dev": {
         "amphp/artax": "^2",
         "fabpot/php-cs-fixer": "^1.9",
         "http2jp/hpack-test-case": "1.0",
-        "phpunit/phpunit": "^5.2"
+        "phpunit/phpunit": "^5.2",
+        "nikic/fast-route": "^1"
     },
     "autoload": {
         "psr-4": {


kelunik:
Suggested packages are horrible, because they don't allow defining compatible version constraints. I'd rather move the router out to a separate package than moving the dependency into "suggest".

bwoebi:
Is there any win in not requiring FastRoute?
I think the Router is a pretty central feature, I'd dislike to move out. Sure, you may use your own router, but it'll probably be used by many.

tyx:
IMO any current application get already a router. (this our case). So we want to use aerys but we will keep routing stuff in our side.

kelunik:
Using your own router is perfectly fine, why do we need to make FastRoute optional for that?

tyx:
Because I like to install only what I really need.

But not a big deal, I was just thinking it was a small mistake so here is the PR. I will use a composer workaround to avoid to install fastroute.

kelunik:
@bwoebi What do you think about moving the router to a separate package?

staabm:
wouldnt it be a single file repos?

kelunik:
@staabm Basically, yes. But only due to `FastRoute` containing everything that's needed already.

rdlowrey:
I'm personally in favor of splitting out the router into its own component library.

bwoebi:
I personally consider it a _deterrent_ when I have to first find a shitton of mircocomponents to even start at all…

bwoebi:
Never heard of that; I personally never use these composer commands though, but directly edit the json file always.
In this case it may be acceptable though.

kelunik:
> I personally consider it a deterrent when I have to first find a shitton of mircocomponents to even start at all…

Everything not having external dependencies is fine. I wouldn't split the document root handler out, as that doesn't have any external dependencies on its own.

> But not a big deal, I was just thinking it was a small mistake so here is the PR. I will use a composer workaround to avoid to install fastroute.

@tyx Could you share your workaround? I actually want something similar to exclude all the polyfills in packages that require PHP 7 or even 7.1.

tyx:
In your composer.json:

```
"replace": {
        "symfony/polyfill-php56": "*",
        "symfony/polyfill-php70": "*"
}
``` 

I don't know if it will continue to work in the next composer release, I hope so as it is very useful currently !

kelunik:
Didn't know about `"replace"`, thanks! Works like a charm.

```json
"replace": {
    "symfony/polyfill-util": "*",
    "symfony/polyfill-php56": "*",
    "paragonie/random_compat": "*"
}
```

PR #137, opened 2017-03-08
Add PHP 7.1 to Travis configuration

joshdifabio:
I just hit an issue where Aerys was not running properly on 7.1.2 but downgrading to 7.1.1 resolved the issue. I've just seen that Travis is only building against 7.0, not 7.1, so this PR adds 7.1 to the Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ language: php
 
 php:
   - 7.0
+  - 7.1
   - nightly
 
 matrix:


bwoebi:
There is a bug in 7.1.2 which will be fixed in 7.1.3. Can't do too much about it.
Will merge this when travis builds will run against 7.1.3.

kelunik:
@joshdifabio You can use 7.1.3RC1, it already contains the fix.

joshdifabio:
Thanks guys. What's the bug?

bwoebi:
There was a possible out-of-bounds access due to not enough memory being allocated for generator stack frames.

joshdifabio:
> There was a possible out-of-bounds access due to not enough memory being allocated for generator stack frames.

Over my head. Thanks though!

PR #141, opened 2017-04-12
Check for NULLs before Loop::disable and Loop::cancel

ck99:
Loop::cancel and Loop::disable now typehint for a string identifier, so all calls need to be checked

### DIFF
--- a/lib/CommandClient.php
+++ b/lib/CommandClient.php
@@ -58,7 +58,9 @@ class CommandClient {
         $bytes = @fwrite($socket, $this->buf);
         if ($bytes == 0) {
             if (!is_resource($socket) || @feof($socket)) {
-                Loop::cancel($this->writeWatcher);
+                if($this->writeWatcher) {
+                    Loop::cancel($this->writeWatcher);
+                }
                 $this->sock = $this->writeWatcher = null;
                 $this->establish();
             }
@@ -66,7 +68,9 @@ class CommandClient {
         }
 
         if ($bytes === \strlen($this->buf)) {
-            Loop::cancel($watcherId);
+            if($watcherId) {
+                Loop::cancel($watcherId);
+            }
             $this->writeWatcher = null;
         }
         $this->written += $bytes;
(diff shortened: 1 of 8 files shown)


bwoebi:
Uh, that's the wrong PR. 
The proper solution would be just removing the assignments of `null`.

lib/CommandClient.php
```diff
@@ -58,15 +58,19 @@ private function writer(string $watcherId, $socket) {
...
+                }
                 $this->sock = $this->writeWatcher = null;
                 $this->establish();
             }
             return;
         }
 
         if ($bytes === \strlen($this->buf)) {
-            Loop::cancel($watcherId);
+            if($watcherId) {
```
bwoebi: this won't ever be null either

lib/CommandClient.php
```diff
@@ -58,15 +58,19 @@ private function writer(string $watcherId, $socket) {
         $bytes = @fwrite($socket, $this->buf);
         if ($bytes == 0) {
             if (!is_resource($socket) || @feof($socket)) {
-                Loop::cancel($this->writeWatcher);
+                if($this->writeWatcher) {
+                    Loop::cancel($this->writeWatcher);
```
bwoebi: better just use $watcherId here then

lib/Websocket/Rfc6455Gateway.php
```diff
@@ -231,7 +231,9 @@ public function reapClient(string $watcherId, InternalRequest $ireq) {
             $client->writeDeferred = new Deferred; // dummy to prevent error
             $client->framesSent = -1;
         } else {
-            Loop::disable($client->writeWatcher);
+            if($client->writeWatcher) {
```
bwoebi: Superfluous to check here, Loop::onWritable will always return string

lib/Websocket/Rfc6455Gateway.php
```diff
@@ -435,7 +437,7 @@ public function onReadable($watcherId, $socket, Rfc6455Client $client) {
             $client->capacity -= \strlen($data);
             if ($client->capacity < $this->maxBytesPerMinute / 2) {
                 $this->lowCapacityClients[$client->id] = $client;
```
bwoebi: Also here, $watcherId is always a string.

bwoebi:
I'm going to apply the proper changes myself in a few minutes.

bwoebi:
Committed https://github.com/amphp/aerys/commit/8b53917e78949d6749c65297d68a4d6555ade309 :-)

PR #145, opened 2017-05-10
Add possibility to specify php options for workers

torinaki:
There must be possibility to adjust workers PHP arguments as well.
I propose following approach:
```
./aerys/vendor/bin/aerys -c ./aerys/server.php --worker-args="-n"
```

### DIFF
--- a/lib/WatcherProcess.php
+++ b/lib/WatcherProcess.php
@@ -388,6 +388,10 @@ class WatcherProcess extends Process {
     private function generateWorkerCommand(Console $console): string {
         $parts[] = \PHP_BINARY;
 
+        if (false === php_ini_scanned_files()) {
+            $parts[] = "-n";
+        }
+
         if ($ini = \get_cfg_var("cfg_file_path")) {
             $parts[] = "-c";
             $parts[] = $ini;


lib/WatcherProcess.php
```diff
@@ -393,6 +393,10 @@ private function generateWorkerCommand(Console $console): string {
             $parts[] = $ini;
         }
 
+        if ($console->isArgDefined("worker-args")) {
+            $parts = array_merge($parts, explode(' ', $console->getArg("worker-args")));
```
kelunik: I'm not sure whether a simple `explode(' ', ...)` is the best way to go here. Does CLImate support one argument to appear multiple times? I don't think so, but that would be much better.

torinaki: Agree, not good approach. It will fail on any space in argument value. The only possible solution to parse it with Regexp. Seems CLImate doesn't support multiple argument. I can access to $argv directly, but it is dirty hack.

Do you have any plans to migrate to Symfony Console component?
It provides possibility to define multiple arguments: InputOption::VALUE_IS_ARRAY

bwoebi:
I'd prefer using Argument\Manager::trailing() for worker args - I mean, what else should we use that for? ;-)

torinaki:
@bwoebi there are two possible types of arguments can be pass to workers: php arguments and climate arguments. I my case I need only php arguments. I'm not sure that trailing args is best choose. It would be better to use trailing args for worker's climate arguments.

bwoebi:
Oh, in that case it would be better to pass `-n` transparently forward. (i.e. if aerys bin has `-n`, also pass `-n` to worker). Same for `-e` and `-z` I guess.
Or is there any reason why one would _just_ want to have the worker running with these options, but not the watcher?

torinaki:
@bwoebi, @kelunik  there no way how to determine which option was passed into PHP. I have removed my changes and did support only for "-n"

PR #148, opened 2017-05-17
Fixing broken tests

ck99:


bwoebi:
Thanks a lot, that's exactly what I've planned to do today :-D
Merging as soon as travis build finishes :-)

ck99:
;)

nice!

lib/Websocket/Message.php
```diff
@@ -2,11 +2,13 @@
...
 namespace Aerys\Websocket;
 
+use Amp\ByteStream\IteratorStream;
+
 class Message extends \Amp\ByteStream\Message {
     private $binary = false;
 
     public function __construct(\Amp\Iterator $iterator, bool $binary = false) {
-        parent::__construct($iterator);
+        parent::__construct(new IteratorStream($iterator));
```
kelunik: I guess `Message` should just accept an `InputStream` and the `new IteratorStream` should be done when constructing the object?

bwoebi: [That's the type of issue we get by having separate Iterator and InputStream…]
Also, I don't care at that level. The message is only instantiated within Rfc6455Gateway and meh.

PR #155, opened 2017-07-20
Move socket transfer to CommandClient

bwoebi:
PR for commenting

### DIFF
--- a/lib/CommandClient.php
+++ b/lib/CommandClient.php
@@ -7,6 +7,7 @@ use Amp\Coroutine;
 use Amp\Deferred;
 use Amp\File;
 use Amp\Loop;
+use Amp\Parser\Parser;
 use Amp\Promise;
 use Amp\Socket;
 use function Amp\call;
@@ -18,10 +19,14 @@ class CommandClient {
     /** @var \Amp\Socket\ClientSocket */
     private $socket;
     private $socketPromise;
-    private $readDeferreds = [];
+    private $readMessages = [];
+
+    /** @var \Amp\Parser\Parser */
+    private $parser;
 
     public function __construct(string $config) {
         $this->path = self::socketPath($config);
+        $this->parser = new Parser(self::parser($this->readMessages));
     }
 
     public static function socketPath(string $config) {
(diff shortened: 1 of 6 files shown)


lib/CommandClient.php
```diff
@@ -26,22 +32,82 @@ public static function socketPath(string $config) {
...
+                $message = \json_encode($message) . "\n";
+                yield $this->socket->write($message);
+            } catch (ClosedException $e) {
+                yield from $this->establish();
+                yield $this->socket->write($message);
+            }
+
+            $promise = ($this->readDeferreds[] = new Deferred)->promise();
+
+            // we can ONLY resolve the socketDeferred AFTER the first write, otherwise order of send()'s is not respected
```
kelunik: Do we really need that? Registered `onResolve` callbacks should be executed in order.

lib/CommandClient.php
```diff
@@ -26,22 +32,82 @@ public static function socketPath(string $config) {
...
+            $parser->push($chunk);
+        }
+
+        $socket = null;
+        $deferreds = $readDeferreds;
+        $readDeferreds = [];
+        foreach ($deferreds as $deferred) {
+            $deferred->fail(new ClosedException("Connection went away ..."));
+        }
+    }
```
kelunik: Maybe use https://github.com/kelunik/streaming-json? It's not too complex here, but why not reuse where possible.

lib/CommandClient.php
```diff
@@ -26,22 +32,82 @@ public static function socketPath(string $config) {
...
+            yield $this->socketDeferred;
+            return;
+        }
+
+        $this->socketDeferred = new Deferred;
+
         $unix = \in_array("unix", \stream_get_transports(), true);
         if ($unix) {
-            return yield Socket\connect("unix://$this->path.sock");
+            $uri = "unix://$this->path.sock";
```
kelunik: Why is the `.sock` added here instead of where `$this->path` is set?

lib/CommandClient.php
bwoebi: I wish to not use packages which depend on amp within amphp.

Also, it really doesn't save much … like a 3 lines.

lib/CommandClient.php
bwoebi: I agree on that, but will fix that in an independent commit then.

lib/CommandClient.php
kelunik: Why not?

bwoebi: I wish to be able to change amp and all its dependents within amphp without depending on external libraries to update first.

It's about flexibility.

kelunik: It's not a random external library. We use `LibDNS`, too.

bwoebi: Which doesn't depend on amp. Just like climate too.

kelunik: It's a small package, doesn't need maintenance, and is maintained by one of the core team (me), I don't see that problem.

bwoebi: Assuming that you will continue to maintain amp forever. Assuming that we don't release dependency upgrades causing small maintenance.

… I'm just not going to buy these assumptions, for any package within amp. I'd like to be consequent here and just won't do that.

If we make that class part of ByteStream package for example, I'm totally fine with using it though.

kelunik: I don't see why it should be part of the byte-stream package directly.

> … I'm just not going to buy these assumptions, for any package within amp. I'd like to be consequent here and just won't do that.

It can always be forked or transferred later should that become true.

lib/CommandClient.php
```diff
@@ -26,22 +32,77 @@ public static function socketPath(string $config) {
...
+
+                yield $this->socketPromise;
+            }
+
+            try {
+                $message = \json_encode($message) . "\n";
+                yield $this->socket->write($message);
+            } catch (ClosedException $e) {
+                $this->socket = null;
+                new Coroutine(self::start($this->path, $this->socket, $this->socketDeferred, $this->readDeferreds));
```
trowski: `$this->socketDeferred` should be `$this->socketPromise`

lib/CommandClient.php
```diff
@@ -26,22 +32,77 @@ public static function socketPath(string $config) {
...
-            return yield $this->socket->write($message);
+            $promise = ($this->readDeferreds[] = new Deferred)->promise();
+
+
+            return $promise;
         });
     }
 
-    private function establish(): \Generator {
+    private static function start($path, &$socket, &$socketPromise, &$readDeferreds) {
```
trowski: This function seems very awkward using things by reference. I'd like to suggest another approach that I can put together later today.

bwoebi: it's about having the destructor trigger if the object is not in use anymore.

lib/CommandClient.php
bwoebi: uh, right. Too much renaming.

lib/CommandClient.php
```diff
@@ -26,22 +32,77 @@ public static function socketPath(string $config) {
...
+                    new Coroutine(self::start($this->path, $this->socket, $this->socketPromise, $this->readDeferreds));
+                }
+
+                yield $this->socketPromise;
+            }
+
+            try {
+                $message = \json_encode($message) . "\n";
+                yield $this->socket->write($message);
+            } catch (ClosedException $e) {
```
trowski: What's going on here? Why catch this exception and try again?

bwoebi: Well, if the connection has been severed in the time since the last command?

trowski: The connection to the watcher process? If that happens I doubt the retry would work or is there something I'm not understanding?

bwoebi: hmm, you should actually be right as it's anyway a local socket. Will remove that part.

lib/CommandClient.php
```diff
@@ -114,33 +96,20 @@ public function stop(): Promise {
     }
 
     public function importServerSockets($addrCtxMap): Promise {
-        return \Amp\call(function() use ($addrCtxMap) {
+        return call(function() use ($addrCtxMap) {
             $reply = yield $this->send(["action" => "import-sockets", "addrCtxMap" => array_map(function ($context) { return $context["socket"]; }, $addrCtxMap)]);
 
-            // replace $this->socket temporarily by a dummy read() which will later be resolved to $this->socket->read() in order to avoid simultaneous read watchers here
```
bwoebi: hmm, can that be removed? won't `$commandClient->imprtServerSockets(…); $commandClient->stop();` - just each immediately after the other cause problems? If there's an active read watcher from the next command?

Issue #124, opened 2017-01-12
Disable host / port match for single host applications


lukaszwit:
Is there any way to use aerys server without host name (by ip). 

```php
(new Aerys\Host)
    ->expose('0.0.0.0', 8089)
    // ... ->use
```
In this case aerys will assign "localhost" name by default but I need to access to it (API) without any host name just by some IP and port. When I'm trying to access it I get an error:
```html
<html>
<body>
<h1>400 Bad Request</h1>

<hr/>
<em>Thu, 12 Jan 2017 17:17:02 GMT</em>
<br/><br/>
</body>
</html>
```

My server logs looks like this:
```
2017-01-12T17:11:19.508998817Z [2017-01-12 17:11:19] debug accept 172.17.0.1:53013
2017-01-12T17:11:19.509241443Z [2017-01-12 17:11:19] debug GET /status HTTP/1.1 @ 172.17.0.1:53013
2017-01-12T17:11:19.509781032Z [2017-01-12 17:11:19] debug close 172.17.0.1:53013
```
Any hint?

kelunik:
@lukaszwit This doesn't happen any longer with the latest changes. Dunno if there's a release yet.

kelunik:
See https://github.com/amphp/aerys/commit/25ebac2702411db8e4db7de4b80609c94739700a

bwoebi:
Or basically, update your Aerys version to 0.4.4.

lukaszwit:
@bwoebi I'm already on 0.4.4
This is my composer.lock:
```json
            "name": "amphp/aerys",
            "version": "v0.4.4",
            "source": {
                "type": "git",
                "url": "https://github.com/amphp/aerys.git",
                "reference": "de8bc2f93bd1407c5762ac7ed88ebf14b5b9add7"
            }
```

Setting empty name also doesn't resolve problem
```php
(new Aerys\Host)
->name('')
->expose('0.0.0.0', 8089)
;
```

When I put the name and run locally it runs perfectly.

kelunik:
@lukaszwit Works fine for me, both on `http://localhost:1337/` and on `http://test.localhost:1337/` with `bin/aerys -c demo.php`.

bwoebi:
@lukaszwit I cannot reproduce your issue either :-(

Could you please try to debug a bit yourself?
The host selection happens in `Aerys\VhostContainer`.
With that configuration `isset($this->vhosts[$wildcardHost])` should return true.

[...]

PR #170, opened 2017-09-04
Add option to skip host matching for single vhost servers

trowski:
Addresses #124.

If the server contains only a single VHost and the `strictHostMatch` option is set to `false`, the single VHost is always selected regardless of the Host header.

### DIFF
--- a/lib/VhostContainer.php
+++ b/lib/VhostContainer.php
@@ -10,9 +10,11 @@ class VhostContainer implements \Countable, Monitor {
     private $defaultHttpDriver;
     private $setupHttpDrivers = [];
     private $setupArgs;
+    private $options;
 
-    public function __construct(HttpDriver $driver) {
+    public function __construct(HttpDriver $driver, Options $options = null) {
         $this->defaultHttpDriver = $driver;
+        $this->options = $options;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lib/VhostContainer.php
```diff
@@ -10,9 +10,11 @@ class VhostContainer implements \Countable, Monitor {
     private $defaultHttpDriver;
     private $setupHttpDrivers = [];
     private $setupArgs;
+    private $options;
 
-    public function __construct(HttpDriver $driver) {
+    public function __construct(HttpDriver $driver, Options $options = null) {
         $this->defaultHttpDriver = $driver;
+        $this->options = $options;
```
bwoebi: You ought using `$ireq->client->options` instead of dependency injecting it here.

lib/VhostContainer.php
```diff
@@ -150,37 +164,42 @@ public function getDefaultHost(): Vhost {
 
     private function selectHostByAuthority(InternalRequest $ireq) {
         $explicitHostId = "{$ireq->uriHost}:{$ireq->uriPort}";
+
+        if (isset($this->vhosts[$explicitHostId])) {
+            return $this->vhosts[$explicitHostId];
+        }
+
+        if ($this->cachedVhostCount === 1 && isset($this->options) && !$this->options->strictHostMatch) {
```
bwoebi: I'm not so sure whether we want to just unconditionally allow it on any single vhost. As far as I see, it actually only makes sense on actual wildcard hosts... (If your host has a name, just remove the name then.)
It looks to me like unconditionally allowing it on non-wildcard hostnames is a bad idea - either you want a name _or_ you want it unconditionally.

trowski: Matching unconditionally requires having only a single host as well as setting an option, something seemingly difficult to do unintentionally. However I agree – this should really only work for wildcard hosts.

Is there an easy way to check if the single host is a wildcard host or will I have to set up another property or two in the container to keep track of wildcard hosts?

lib/VhostContainer.php
trowski: Sure, simple enough.

lib/VhostContainer.php
bwoebi: There are only two wildcard host variants as you see a few lines below, they start with either `0.0.0.0:` or `[::]:`.

bwoebi:
Another option instead of having an extra option for this would be allowing something like `":0"` or `"foo.bar:0"` (alongside with `":80"` or `"foo.bar:80"`) within `Host->name()` which explicitly tells the server to port-match against a specific given port, not necessarily equal to the port we're actually listening on, or not at all.

I have a preference for that as it keeps it very explicit and local to the actual `Host` definition and avoids adding an extra option. Additionally it does allow to _not_ skip name matching, but only port matching then (when using `foo.bar:0`).
Additionally it enables you to use two (or multiple) different hosts (or domains) which are all proxied - the current design does not allow managing two domains like `foo.bar`  and `baz.bar` on the same Aerys instance when they ought both being proxied.

lib/VhostContainer.php
kelunik: @bwoebi What's with wildcard hosts listening on `127.0.0.1` then?

kelunik:
We can do that, too, but with `"*:*"` or `"name:*"` instead of `0`.

bwoebi:
I do not care too much whether we'll call it 0 or * or even allow both. Whatever is fine here will work.

bwoebi:
See #172 for an implementation of that.

PR #172, opened 2017-09-07
Fix #124 (the second)

### LINKED ISSUE #124: Disable host / port match for single host applications

bwoebi:
An alternate solution for #124 (previous PR at #170):

You can now set a specific expected port the client must specify in the Host header or wildcard port ("*") to avoid any port matching there.

### DIFF
--- a/test/VhostTest.php
+++ b/test/VhostTest.php
@@ -9,36 +9,73 @@ use Aerys\VhostContainer;
 use PHPUnit\Framework\TestCase;
 
 class VhostTest extends TestCase {
-    public function testVhostSelection() {
-        $vhosts = new VhostContainer($this->createMock('Aerys\HttpDriver'));
-        $localvhost = new Vhost("localhost", [["127.0.0.1", 80], ["::", 8080]], function () {}, [function () {yield;}]);
-        $vhosts->use($localvhost);
-        $vhost = new Vhost("", [["127.0.0.1", 80], ["::", 80]], function () {}, [function () {yield;}]);
-        $vhosts->use($vhost);
-
-        $this->assertEquals(2, $vhosts->count());
-
-        $ireq = new InternalRequest;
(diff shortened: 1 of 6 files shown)


test/VhostTest.php
```diff
@@ -64,28 +101,28 @@ public function testCryptoVhost() {
...
      */
     public function testBadInterface() {
-        new Vhost("", [["rd.lo.wr.ey", 1025]], function () {}, []);
+        new Vhost("*", [["rd.lo.wr.ey", 1025]], function () {}, []);
     }
 
     /**
      * @expectedException \LogicException
-     * @expectedExceptionMessage Cannot have two hosts with the same `localhost:80` name
+     * @expectedExceptionMessage Cannot have two hosts with the same `localhost` name on the `127.0.0.1:80` interface
```
kelunik: It's still weird, should be `same name (...) on the same interface (...)` IMO.

Issue #168, opened 2017-08-31
Support listening on Unix sockets


kelunik:
Currently we only support listening on IPs, but not on Unix sockets. Might be interesting if Aerys is deployed behind Nginx or another reverse proxy.

bwoebi:
I generally like the idea.

I am wondering though how this will work with multiple binding. SO_REUSEPORT won't work here.
So only option is to force usage of socket transfer then.

Additionally, how will the signature on Host::expose() look like then? `Host::expose(string $address, int $port = 0)` and require the port being non-zero for anything not being a path (and zero for paths)?

That way it should be pretty trivial, except that we have to special case it within CommandClient::importServerSockets (in order to use so_reuseport when available on tcp sockets, but not on unix sockets - ultimately so_reuseport is the fastest way to use).

kelunik:
Either `0` or `null` for the port, yes.

umbri:
https://github.com/amphp/uri/issues/9

Please don't forget about this

kelunik:
@umbri That's a separate issue.

PR #173, opened 2017-09-08
Unix Domain Socket support (PR for #168)

bwoebi:
As noted in #168: it is necessary to always use socket transfer for Unix Domain Sockets, thus we need to decide ourselves in worker whether to use socket transfer or not.

Thus, unix domain sockets should generally be impossible with < PHP 7.0.7 and without sockets extension in production mode (i.e. when using multiple workers).

### DIFF
--- a/lib/Host.php
+++ b/lib/Host.php
@@ -10,11 +10,11 @@ class Host {
     private $httpDriver;
 
     /**
-     * Assign the IP and port on which to listen.
+     * Assign the IP or unix domain socket and port on which to listen.
      *
-     * The address may be any valid IPv4 or IPv6 address. The "0.0.0.0" indicates
-     * "all IPv4 interfaces" and is appropriate for most users. Use "::" to indicate "all IPv6
-     * interfaces". To indicate "all IPv4 *and* IPv6 interfaces", use a "*" wildcard character.
+     * The address may be any valid IPv4 or IPv6 address or unix domain socket path. The "0.0.0.0"
+     * indicates "all IPv4 interfaces" and is appropriate for most users. Use "::" to indicate "all
+     * IPv6 interfaces". Use a "*" wildcard character to indicate "all IPv4 *and* IPv6 interfaces".
      *
      * Note that "::" may also listen on some systems on IPv4 interfaces. PHP did not expose the
      * IPV6_V6ONLY constant before PHP 7.0.1.
@@ -24,14 +24,22 @@ class Host {
      * access on UNIX-like systems. The default port for encrypted sockets (https) is 443. If you
      * plan to use encryption with this host you'll generally want to use port 443.
      *
-     * @param string $address The IPv4 or IPv6 interface to listen to
-     * @param int $port The port number on which to listen
+     * @param string $address The IPv4 or IPv6 interface or unix domain socket path to listen to
+     * @param int $port The port number on which to listen (0 for unix domain sockets)
      * @return self
      */
-    public function expose(string $address, int $port): Host {
-        if ($port < 1 || $port > 65535) {
+    public function expose(string $address, int $port = 0): Host {
+        $isPath = $address[0] == "/";
+
+        if ($isPath) {
+            if ($port != 0) {
+                throw new \Error(
+                    "Invalid port number {$port}; must be zero for an unix domain socket"
+                );
+            }
+        } elseif ($port < 1 || $port > 65535) {
             throw new \Error(
-                "Invalid port number; integer in the range 1..65535 required"
+                "Invalid port number {$port}; integer in the range 1..65535 required"
             );
         }
 
@@ -43,9 +51,9 @@ class Host {
             $address = "::";
         }
 
-        if (!@inet_pton($address)) {
+        if (!$isPath && !@inet_pton($address)) {
             throw new \Error(
-                "Invalid IP address"
+                "Invalid IP address or unix domain socket path"
             );
         }
 
--- a/lib/Server.php
+++ b/lib/Server.php
@@ -80,6 +80,7 @@ class Server implements Monitor {
     // private callables that we pass to external code //
     private $exporter;
     private $onAcceptable;
+    private $onUnixSocketAcceptable;
     private $negotiateCrypto;
     private $onReadable;
     private $onWritable;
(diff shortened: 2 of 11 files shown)


lib/Host.php
```diff
@@ -57,14 +65,18 @@ public function expose(string $address, int $port): Host {
     /**
      * Assign a domain name (e.g. localhost or mysite.com or subdomain.mysite.com).
      *
-     * A host name is only required if a server exposes more than one host. If a name is not defined
-     * the server will default to "localhost"
+     * An explicit host name is only required if a server exposes more than one host. If a name is
+     * not defined (or "*") the server will allow any hostname.
```
kelunik: That's not true, no? If two different ports are used, no names are required?

lib/Host.php
```diff
@@ -57,14 +65,18 @@ public function expose(string $address, int $port): Host {
...
+     *
+     * By default the port must match with the interface. It is possible to explicitly require
+     * a specific port in the hostname by appending ":port" (e.g. "localhost:8080"). It also here is
+     * possible to specify a wildcard with "*" (e.g. "*:*" to accept any hostname from any port).
```
kelunik: _It also here is possible_ → It is also possible_

lib/Server.php
```diff
@@ -245,7 +242,18 @@ private function doStart(callable $bindSockets): \Generator {
...
+                @unlink(substr($address, 7));
+            }
+
+            if (!$socket = stream_socket_server($address, $errno, $errstr, STREAM_SERVER_BIND | STREAM_SERVER_LISTEN, stream_context_create($context))) {
+                throw new \RuntimeException(sprintf("Failed binding socket on %s: [Err# %s] %s", $address, $errno, $errstr));
+            }
+
+            return $socket;
+        };
+        $this->boundServers = yield $bindSockets($this->generateBindableAddressContextMap(), $socketBinder);
```
kelunik: This should probably be `$this->boundSockets`?

lib/Server.php
```diff
@@ -323,7 +335,7 @@ private function onAcceptable(string $watcherId, $server) {
             return;
         }
 
-        \assert($this->logDebug("accept {$peerName}"));
+        \assert($this->logDebug("accept {$peerName} #" . (int) $client));
```
kelunik: Should probably be `"Accepting TCP connection from {$peerName} #" . (int)  $client`

lib/Server.php
```diff
@@ -336,6 +348,17 @@ private function onAcceptable(string $watcherId, $server) {
         }
     }
 
+    private function onUnixSocketAcceptable(string $watcherId, $server) {
+        if (!$client = @\stream_socket_accept($server, $timeout = 0)) {
+            return;
+        }
+
+        \assert($this->logDebug("accept connection on unix socket #" . (int) $client));
```
kelunik: Should probably be `"Accepted Unix connection #" . (int)  $client`

lib/VhostContainer.php
```diff
@@ -110,81 +111,73 @@ public function setupHttpDrivers(...$args) {
...
         }
 
-        // IMPORTANT: Wildcard IP hosts without names that are running both encrypted and plaintext
-        // apps on the same interface (via separate ports) must be checked for encryption to avoid
-        // displaying unencrypted data as a result of carefully crafted Host headers. This is an
-        // extreme edge case but it's potentially exploitable without this check.
-        // DO NOT REMOVE THIS UNLESS YOU'RE SURE YOU KNOW WHAT YOU'RE DOING.
-        if ($vhost->isEncrypted() != $ireq->client->isEncrypted) {
-            return null;
+        $addressPortWildcardHost = "*:0{$serverId}";
```
kelunik: Why that `0` in there?

lib/Server.php
bwoebi: if you want to be precise you could call it boundServerSockets - but really doesn't matter. I just retained the existing name.

lib/Host.php
bwoebi: \* on a same interface. Yeah, that's missing.

lib/WorkerProcess.php
```diff
@@ -31,9 +30,33 @@ protected function doStart(Console $console): \Generator {
...
-            \assert(\extension_loaded("sockets") && PHP_VERSION_ID > 70007);
-            yield $server->start([new CommandClient((string) $console->getArg("config")), "importServerSockets"]);
+        if (\extension_loaded("sockets") && PHP_VERSION_ID > 70007 && stripos(PHP_OS, "WIN") !== 0) {
+            // needs socket_export_stream() and no Windows
+            $config = (string) $console->getArg("config");
+            $getSocketTransferImporter = function () use ($config, &$socketTransfer) {
+                return $socketTransfer ?? $socketTransfer = [new CommandClient($config), "importServerSockets"];
+            };
+
+            if (in_array(strtolower(PHP_OS), ["darwin", "freebsd"])) { // the OS not supporting *round-robin* so_reuseport client distribution
```
kelunik: Is that really `PHP_OS` or rather `PHP_OS_FAMILY`?

test/ClientTest.php
```diff
@@ -121,8 +121,9 @@ public function testClientDisconnect() {
 
             $context = (new ClientTlsContext)->withoutPeerVerification();
             $client = new DefaultClient(null, null, $context);
+            $port = parse_url($address, PHP_URL_PORT);
             $promise = $client->request(
-                (new \Amp\Artax\Request("https://$address/", "POST"))->withBody("body")
+                (new \Amp\Artax\Request("https://localhost:$port/", "POST"))->withBody("body")
```
kelunik: Why that?

lib/VhostContainer.php
bwoebi: Allow any port.

test/ServerTest.php
```diff
@@ -333,8 +337,15 @@ protected function output(string $message) { /* /dev/null */
         return [$address, $server];
     }
 
-    public function testUnencryptedIO() {
-        Loop::run(function () {
+    public function provideFalseTrueUnixDomainSocket() {
```
kelunik: WAT :D

lib/WorkerProcess.php
bwoebi: PHP_OS_FAMILY only has been introduced recently. For now PHP_OS must suffice.

test/ClientTest.php
bwoebi: Because of the single vhost == default host not holding anymore and being explicit now.

lib/VhostContainer.php
kelunik: Should probably be `*` then as in the definition of hosts?

lib/WorkerProcess.php
kelunik: See https://github.com/symfony/polyfill-php72 for a polyfill.

lib/VhostContainer.php
bwoebi: Nah, makes things easier to keep this as integer. It's anyway just an implementation detail between Vhost and VhostContainer mostly.

kelunik: They should probably both be marked as `@internal`.

lib/WorkerProcess.php
bwoebi: Also, OpenBSD actually supports round-robin so_reuseport for example. Even though it is listed in the same family as FreeBSD (bsd). So, PHP_OS is definitely correct here.

PR #177, opened 2017-09-15
Remove magic __call from Router

### LINKED ISSUE #159: Remove magic from Router
Router currently converts any non-existent method to `route(strtoupper($method, ...$args))`. This should be removed, as any new method on `Router` is a potential BC break with that. Instead, people should just use `route("GET", ...)` instead of `get(...)`.

If we want, we can add shortcuts for the common methods, but I don't see too much advantage in that.

kelunik:
Fixes #159.

### DIFF
--- a/demo.php
+++ b/demo.php
@@ -18,50 +18,50 @@ const AERYS_OPTIONS = [
 /* --- http://localhost:1337/ ------------------------------------------------------------------- */
 
 $router = router()
-    ->get("/", function(Request $req, Response $res) {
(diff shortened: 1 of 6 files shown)


bwoebi:
Please also update demo.php and the docs.

kelunik:
Already noticed that, yes. Updated docs, `demo.php` and tests.

bwoebi:
LGTM. Let's also see what @trowski says.

PR #185, opened 2017-09-28
Add -u --user option to the command line api

### LINKED ISSUE #140: Add --user command line option to aerys
AFAIK, and correct me if this is wrong, it's only possible to pass the user, from which workers will run, through a constant called `AERYS_OPTIONS`. Ex.:

```php
const
    AERYS_OPTIONS = [
        "user" => "marcio",
        "defaultContentType" => "application/json",
        ...
    ]
;
``` 

This is fine if you run from the same unprivileged user on all your dev and prod servers. But that can be a PITA if you need to set different users for different servers - because we obviously can not `getenv("MY_API_USER")` on a constant declaration in PHP.

I'd like to suggest a `-u` and `--user` option to aerys command line interface for that purpose. The rational is simply convenience - at least I noticed the "user" option is the only one that will certainly vary and can't be getenved like the exposed port.

Thoughts?

maxmeijer42:
Closes #140

### DIFF
--- a/lib/Internal/functions.php
+++ b/lib/Internal/functions.php
@@ -124,6 +124,9 @@ function bootServer(PsrLogger $logger, Console $console): \Generator {
     }
 
     $options["debug"] = $console->isArgDefined("debug");
+    if($console->isArgDefined("user")){
+        $options["user"] = $console->getArg("user");
+    }
     $options["configPath"] = $configFile;
 
     return initServer($logger, $hosts, $options);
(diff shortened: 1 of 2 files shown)


lib/Internal/functions.php
```diff
@@ -124,6 +124,9 @@ function bootServer(PsrLogger $logger, Console $console): \Generator {
     }
 
     $options["debug"] = $console->isArgDefined("debug");
+    if($console->isArgDefined("user")){
```
bwoebi: whitespace after if keyword and before {.

PR #187, opened 2017-10-01
Throw descriptive error if no hosts are returned from the config file

### LINKED ISSUE #157: Useful error message for no hosts returned from config file
If no hosts are returned from the config file, Aerys currently fails with `[2017-07-25 12:05:12] critical TypeError: Expected an array of Hosts as second parameter, but array also contains integer in /home/kelunik/GitHub/kelunik/chat/vendor/amphp/aerys/lib/functions.php:541`, which is not really a helpful error message.

maxmeijer42:
Closes #157

### DIFF
--- a/lib/Internal/functions.php
+++ b/lib/Internal/functions.php
@@ -108,6 +108,12 @@ function bootServer(PsrLogger $logger, Console $console): \Generator {
         $hosts = [$hosts];
     }
 
+    if (empty($hosts)) {
+        throw new \Error(
+            "Config file at $configFile did not return any hosts"
+        );
+    }
+
     if (!defined("AERYS_OPTIONS")) {
         $options = [];
     } elseif (is_array(AERYS_OPTIONS)) {


lib/Internal/functions.php
```diff
@@ -108,6 +108,12 @@ function bootServer(PsrLogger $logger, Console $console): \Generator {
         $hosts = [$hosts];
     }
 
+    if (empty($hosts)) {
+        throw new \Error(
+            "Config file at $configFile did not return any hosts"
+        );
+    }
```
kelunik: @bwoebi Should we remove the `if ($hosts === false)` check above then?

bwoebi: include returns false if the file could not be loaded at all. This check here is throwing when the return value of the file is unexpected.

kelunik: Should probably just be `require` then? As it might also return `false` (in theory)?

bwoebi: require will throw an \Error, yes, but it won't tell you that the file trying to be included was the indeed config file and not some random internal file (i.e. a bug on our side).

kelunik: Uhm, the error tells you the file name?

maxmeijer42: IMHO it would indeed be clearer if the error message specifies that it was a config file inclusion that failed (as opposed to some random internal file). Differentiating between the config file returning `false` and not existing, would also be nice. To get both these things working correctly, you could check if the file exists before `require`ing it.

kelunik:
Thanks!

PR #190, opened 2017-10-02
Add warning for privileged run in case when user is set to root via options

### LINKED ISSUE #189: There's no warning when the 'user' option is set to "root"
There's a warning when server starts with `sudo`. But there isn't one when `user` option is set to `"root"` in `AERYS_OPTIONS`.

ztsu:
Closes #189


ztsu:
It's just a warning and doesn't prevent starting the server. Should I change the message for this case?

kelunik:
Don't you think a user who sets it to root knows what he does? Setting it explicitly to root could be a way to silence the warning.

Issue #188, opened 2017-10-01
Generators from config file should handle promise returns correctly


kelunik:
https://github.com/amphp/aerys/blob/7e01ef45b71303a8d5294b3ddbf0ed8f4760397b/lib/Internal/functions.php#L104

This should work if the return value of the generator is a promise. Either use `Amp\call` instead of `yield from` or remove the `else` from the other branch.

krlv:
@kelunik Could you please provide an example where return value of the generator is a promise?
I'm trying to reproduce this use case, but constantly get `NULL` from generator.

kelunik:
Fixed via #191.

PR #191, opened 2017-10-07
Fix generator promise return

krlv:
#188 Fix:

1.  If config is `\Generator` get its return value and continue following checks
2. Update `test/testBootServerInclude.php` to cover this case
3. Rename `BootstrapperTest` -> `bootServerTest` class


kelunik:
Thanks!

PR #193, opened 2017-10-13
GET param values should be decoded with urldecode

brstgt:
E.g. it is valid to have a param like "foo=bar+baz" which should decode to "bar baz" which is not the case with rawurldecode

### DIFF
--- a/lib/StandardRequest.php
+++ b/lib/StandardRequest.php
@@ -120,7 +120,7 @@ class StandardRequest implements Request {
         foreach ($pairs as $pair) {
             $pair = explode("=", $pair, 2);
             // maxFieldLen should not be important here ... if it ever is, create an issue...
-            $this->queryParams[rawurldecode($pair[0])][] = rawurldecode($pair[1] ?? "");
+            $this->queryParams[rawurldecode($pair[0])][] = urldecode($pair[1] ?? "");
         }
 
         return $this->queryParams;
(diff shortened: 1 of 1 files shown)


lib/StandardRequest.php
```diff
@@ -120,7 +120,7 @@ private function parseParams() {
         foreach ($pairs as $pair) {
             $pair = explode("=", $pair, 2);
             // maxFieldLen should not be important here ... if it ever is, create an issue...
-            $this->queryParams[rawurldecode($pair[0])][] = rawurldecode($pair[1] ?? "");
+            $this->queryParams[rawurldecode($pair[0])][] = urldecode($pair[1] ?? "");
```
kelunik: Should the same apply to the name?

brstgt:
I am not 100% sure but i think i remember that name + values have different
rules. Maybe I mix it up and it was the path vs get params.

2017-10-13 11:05 GMT+02:00 Niklas Keller <notifications@github.com>:

bwoebi:
No, name and value are identical. The BodyParser uses urldecode(). Not sure why I put rawurldecode() there two years ago.

Also, yes, path definitely needs rawurldecode(), query string does need urldecode().

brstgt:
... changed ...

kelunik:
See also https://github.com/amphp/uri/issues/12, dunno whether we need a new major for that there.

PR #197, opened 2017-10-23
Add Response::abort() function

### LINKED ISSUE #194: Add possibility to close connection / stream
There should be a possibility to close a connection / stream (for H2) to abort requests if things go wrong. One such example is a reverse proxy, where the remote hangs up before the full response is sent.

**Real World Example**
```php
$response->setHeader("content-type", $proxyResponse->getHeader("content-type") ?? "image/jpeg");
$response->setHeader("cache-control", "public, max-age=31536000");

try {
    yield ByteStream\pipe($proxyResponse->getBody(), $response);
} catch (HttpException $e) {
    // what should I do now?
}
```

bwoebi:
Fixes #194.

### DIFF
--- a/lib/Response.php
+++ b/lib/Response.php
@@ -103,6 +103,14 @@ interface Response extends \Amp\ByteStream\OutputStream {
     public function end(string $finalBodyChunk = ""): \Amp\Promise;
 
     /**
+     * Signal an extraordinary response end.
+     *
+     * It fails an eventual pending request body message and will reset the
+     * active HTTP stream.
+     */
+    public function abort();
+
+    /**
      * Indicate resources which a client likely needs to fetch. (e.g. Link: preload or HTTP/2 Server Push).
      *
      * @param string $url The URL this request should be dispatched to
--- a/lib/StandardResponse.php
+++ b/lib/StandardResponse.php
@@ -143,14 +143,8 @@ class StandardResponse implements Response {
     }
 
     /**
-     * Stream partial entity body data.
-     *
-     * If response output has not yet started headers will also be sent
-     * when this method is invoked.
-     *
-     * @param string $partialBody
+     * {@inheritDoc}
      * @throws \Error If response output already complete
-     * @return \Amp\Promise to be succeeded whenever local buffers aren't full
      */
     public function write(string $partialBody): \Amp\Promise {
         if ($this->state & self::ENDED) {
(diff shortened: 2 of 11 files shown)


lib/Websocket/Handshake.php
```diff
@@ -111,6 +111,13 @@ public function end(string $finalBodyChunk = ""): \Amp\Promise {
     /**
      * {@inheritDoc}
      */
+    public function abort(): \Amp\Promise {
```
kelunik: Why does this have `abort()`? And why does it declare `: Amp\Promise`?

lib/Response.php
```diff
@@ -103,6 +103,14 @@ public function flush();
     public function end(string $finalBodyChunk = ""): \Amp\Promise;
 
     /**
+     * Signal an extraordinary response end.
+     *
+     * It succeeds an eventual pending request body message and will reset the
+     * active HTTP stream.
```
kelunik: Why does it succeed a pending request body message instead of failing it?

lib/Websocket/Handshake.php
bwoebi: It has abort because interface. And the return type declaration is really too much.

lib/Response.php
bwoebi: It actually fails it, the docstring is wrong.

lib/StandardResponse.php
```diff
@@ -205,15 +205,22 @@ public function flush() {
...
+     * the following:
      *
-     *     $response->write($finalBody);
+     *     $response->write($finalBodyChunk);
      *     $response->end();
      *
-     * @param string $finalBody Optional final body data to send
+     * Note: Invoking Response::end() with a non-empty $finalBodyChunk parameter
+     * without having previously invoked Response::write() is equivalent to calling
+     * Response::send($finalBodyChunk).
```
kelunik: `Response::send()` doesn't exist.

bwoebi: I've copied it from the interface, should probably be fixed there.

kelunik: Why did you copy it instead of using `@inheritdoc`?

kelunik: @bwoebi ping

kelunik:
I think we should branch of `0.7.x` before merging it, so we don't end up with what we did with `v0.6.1`.

kelunik:
This is no longer compatible with the approach taken in `0.8.x` returning responses. I guess this should be a method on `Request` instead that, when invoked, makes the server ignore any response and abort the request via RST on either the stream level for HTTP/2 or the TCP level for HTTP/1.

kelunik:
I'm closing this one for the reason mentioned above.

PR #201, opened 2017-10-27
Remove log level recommandation log for production

shouze:
Everyone can log at the level he want in fact.

### DIFF
--- a/lib/WatcherProcess.php
+++ b/lib/WatcherProcess.php
@@ -275,9 +275,9 @@ class WatcherProcess extends Process {
         if (!isset(Logger::LEVELS[$level])) {
             return;
         }
-        if (Logger::LEVELS[$level] > Logger::LEVELS[Logger::WARNING]) {
-            $this->logger->warning(
-                "Running aerys in production with a log level higher than \"warning\" is not " .
+        if (Logger::LEVELS[$level] === Logger::LEVELS[Logger::DEBUG]) {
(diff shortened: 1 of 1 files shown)


shouze:
ping @rdlowrey.

kelunik:
I think we can just lower that to the level one higher than debug.

bwoebi:
Yeah, wouldn't remove that completely and still warn if it's set to debug.

shouze:
Makes senses I change the PR soon in this direction.

shouze:
@kelunik @bwoebi ok done, that was exhausting 👼

lib/WatcherProcess.php
```diff
@@ -275,7 +275,7 @@ protected function recommendLogLevel(Console $console) {
         if (!isset(Logger::LEVELS[$level])) {
             return;
         }
-        if (Logger::LEVELS[$level] > Logger::LEVELS[Logger::WARNING]) {
+        if (Logger::LEVELS[$level] > Logger::LEVELS[Logger::DEBUG]) {
             $this->logger->warning(
                 "Running aerys in production with a log level higher than \"warning\" is not " .
```
nikic: Should probably adjust this `\"warning\"` too :)

shouze: Ho my bad... yes of course... 👼

shouze:
ping @nikic ok fixed 😉

Issue #199, opened 2017-10-26
How to implement a NCSA Common Line Format (CLF) logger middleware?


shouze:
I've tried to procude NCSA CLF compatible logs for some REST Http server managed by aerys, here's the current state of my work:
```php
class NcsaClfLogger implements Aerys\Bootable
{
    private $logger;

    function boot(Aerys\Server $server, Psr\Log\LoggerInterface $logger)
    {
        $this->logger = $logger;
    }

    function __invoke(Aerys\Request $request, Aerys\Response $response)
    {
        // We should probably return and/or partialy collect info until the response is in ended state?
        // NCSA / CLF log line format
        // @see https://en.wikipedia.org/wiki/Common_Log_Format
        $logLine = sprintf(
            '%s %s %s [%s] "%s %s HTTP/%s" %s %s',
            $request->getHeader('X-Forwarded-For') ?? $request->getConnectionInfo()['client_addr'],
            '-',
            '-',
            date('d/M/Y:H:i:s O'),
            $request->getMethod(),
            $request->getUri(),
            $request->getProtocolVersion(),
            $response->__debugInfo()[':status'], // no other way to get the response status + we should ensure it's the final status
            0 // how to get response size? We should probable get it when response is in ended state but how?
        );
        $this->logger->info($logLine);
    }
}
```

And here's the result in aerys logs:
```
aerys -c bin/server
[2017-10-26 13:57:01] warning 203.0.113.195, 70.41.3.18, 150.172.238.178 - - [26/Oct/2017:13:57:01 +0000] "GET /health HTTP/1.1" 200 0
[2017-10-26 13:57:04] warning 172.17.0.1 - - [26/Oct/2017:13:57:04 +0000] "GET /health HTTP/1.1" 200 0
[2017-10-26 13:57:06] warning 203.0.113.195, 70.41.3.18, 150.172.238.178 - - [26/Oct/2017:13:57:06 +0000] "GET /health HTTP/1.1" 200 0
[2017-10-26 13:57:07] warning 172.17.0.1 - - [26/Oct/2017:13:57:07 +0000] "GET /health HTTP/1.1" 200 0
```

I have few questions:
- How to get & return the response size ? Response always seem to be in state `NONE` in the middleware.
- Is there a better way to get the response status and to ensure that's the correct one? Here in my implementation it will always return 200. (also related to response state of course).
- How can I avoid aerys to prefix log lines by itself? Where can I override the default log line format/handler?

kelunik:
[...]

PR #203, opened 2017-11-01
Delay first log line

shouze:
As discussed in https://github.com/amphp/aerys/issues/199#issuecomment-340162206


bwoebi:
I think you're a few hours too late here :-D

https://github.com/amphp/aerys/commit/48964827dffa204a8385e582974013b566f97971#diff-3900e64df972fe6d047baa2446d1dabdR104

Closing as already implemented.

bwoebi:
By the way, I only delay it up to immediately after inclusion, as promise stacktraces may not show the original included file anymore.
But that should be fully sufficient, I think.

shouze:
@bwoebi indeed, so nice & better changeset 👍

PR #205, opened 2017-11-11
Join Allow header values with comma and space (according with RFC 7231)

samizdam:
In most web-servers and http-clients implementations header values joined with comma and space. 
I am fix it for `Allow` header in 2 places.

### DIFF
--- a/lib/Server.php
+++ b/lib/Server.php
@@ -733,7 +733,7 @@ class Server implements Monitor {
         $body = makeGenericBody($status);
         $response->setStatus($status);
         $response->setHeader("Connection", "close");
-        $response->setHeader("Allow", implode(",", $this->options->allowedMethods));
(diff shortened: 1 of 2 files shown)


kelunik:
Just for reference: Actual grammar: `Allow = [ ( "," / method ) *( OWS "," [ OWS method ] ) ]` in https://tools.ietf.org/html/rfc7231#appendix-D

From https://tools.ietf.org/html/rfc7230#section-3.2.3:

> The OWS rule is used where zero or more linear whitespace octets
   might appear.  For protocol elements where optional whitespace is
   preferred to improve readability, a sender SHOULD generate the
   optional whitespace as a single SP; otherwise, a sender SHOULD NOT
   generate optional whitespace except as needed to white out invalid or
   unwanted protocol elements during in-place message filtering.

kelunik:
Thanks!

PR #206, opened 2017-11-11
Use actual server version in constants.

samizdam:


kelunik:
I think we better use something like https://github.com/Ocramius/PackageVersions to determine the version automatically from the used version.

samizdam:
Hm, Ocramius/PackageVersions interesting tool, but get version with hash in composer format like `v0.7.1@74ee552ef7e2d11fe7c3ab4b686c904d67e954c1`. 

I would like actual semver and readable for human value, like [this](https://github.com/ratchetphp/Ratchet/blob/master/src/Ratchet/ConnectionInterface.php#L8). 

Off course I can do it and parse with https://github.com/naneau/semver, but my issue is use this value in x-powered-by header without some magic. 
E.g. I want use some constant from aerys code as is.

kelunik:
You can simply use `strtok($version, "@")` for "parsing".

samizdam:
I amend commit and use ocramius/package-versions.

lib/constants.php
```diff
@@ -3,7 +3,7 @@
 namespace Aerys;
 
 const SERVER_NAME = "aerys";
-const SERVER_VERSION = "0.1.0-dev";
+define('SERVER_VERSION', \PackageVersions\Versions::getVersion('amphp/aerys'));
```
kelunik: I think this needs to define the namespace name in the constant, no?

samizdam: Yes, it is right. 
After fix this code get value `0.6.0.9999999-dev@9fbe753c9962eb6b7adc7ae99a6ef917f6b50a80` on my master branch.
Do you like it?

kelunik:
I've pushed a slightly different version that removes the `999999` for `dev` versions.

PR #207, opened 2017-11-13
Adopt a more portable cpu detection

shouze:
This avoids to have to install nproc for Linux (eg when running aery into a lightweight/minimal docker image). It's also the most portable approach for *NIX systems BTW.

### DIFF
--- a/lib/WatcherProcess.php
+++ b/lib/WatcherProcess.php
@@ -345,13 +345,13 @@ class WatcherProcess extends Process {
                 $cmd = "wmic cpu get NumberOfCores";
                 break;
             case "linux":
-                $cmd = "nproc";
-                break;
-            case "freebsd":
-                $cmd = "sysctl -a | grep 'hw.ncpu' | cut -d ':' -f2";
-                break;
             case "darwin":
-                $cmd = "sysctl -a | grep 'hw.ncpu:' | awk '{ print $2 }'";
+                $cmd = "getconf _NPROCESSORS_ONLN";
+                break;
+            case "netbsd":
+            case "openbsd":
+            case "freebsd":
+                $cmd = "sysctl hw.ncpu | cut -d ':' -f2";
                 break;
             default:
                 $cmd = null;


kelunik:
See also https://groups.google.com/d/msg/comp.unix.shell/nqzUuqsmFo8/eBBR04d-JQwJ.

shouze:
@kelunik ok I've changed only for `Linux`, reverted changes for darwin & freebsd.

shouze:
@kelunik second edit: I've changed accordingly to the link you pointed out. So netbsd & openbsd detection should be ok from now.

lib/WatcherProcess.php
```diff
@@ -345,14 +345,14 @@ private function countCpuCores() {
...
                 break;
             case "linux":
-                $cmd = "nproc";
+            case "darwin":
+                $cmd = "getconf _NPROCESSORS_ONLN";
                 break;
+            case "netbsd":
+            case "openbsd":
             case "freebsd":
                 $cmd = "sysctl -a | grep 'hw.ncpu' | cut -d ':' -f2";
```
kelunik: I guess we could use `sysctl hw.ncpu` here then?

shouze:
@kelunik ok done I've squashed the `*bsd` sysctl call simplification 😉

lib/WatcherProcess.php
```diff
@@ -345,13 +345,13 @@ private function countCpuCores() {
...
+                $cmd = "getconf _NPROCESSORS_ONLN";
                 break;
+            case "netbsd":
+            case "openbsd":
             case "freebsd":
-                $cmd = "sysctl -a | grep 'hw.ncpu' | cut -d ':' -f2";
-                break;
-            case "darwin":
-                $cmd = "sysctl -a | grep 'hw.ncpu:' | awk '{ print $2 }'";
+                $cmd = "sysctl hw.ncpu | cut -d ':' -f2";
```
kelunik: Does it still need the `cut` in that case?

krytarowski: Why not just `getconf NPROCESSORS_ONLN`?

staabm: see https://github.com/amphp/aerys/pull/207#issuecomment-343936901

bwoebi: Yes, output is `hw.ncpu: 8`.

krytarowski: @staabm ?

```
rugged$ getconf NPROCESSORS_ONLN
4
rugged$ uname -a
NetBSD rugged 8.99.1 NetBSD 8.99.1 (GENERIC) #7: Mon Jul  3 15:56:30 CEST 2017  root@chieftec:/public/netbsd-root/sys/arch/amd64/compile/GENERIC amd64
```

kelunik: If it works there, too, we can certainly use it for NetBSD as well. Want to provide a PR?

krytarowski: `sysctl` works too and it's not big gain to change it again, it would be just more elegant to use `getconf` in the first place.

Issue #200, opened 2017-10-27
Change Host::encrypt to use ServerTlsContext


kelunik:
We already have a value object for server TLS in `amphp/socket`, that should be used in Aerys as well instead of relying on arrays.

kelunik:
Closing as it's done in `0.8.x`.

PR #208, opened 2017-11-17
Allow use of ServerTlsContext (or Certificate) within Host::encrypt

bwoebi:
Implements the requested change in #200.

### DIFF
--- a/lib/Host.php
+++ b/lib/Host.php
@@ -2,6 +2,8 @@
 
 namespace Aerys;
 
+use Amp\Socket\{ Certificate, ServerTlsContext };
+
 class Host {
     private $name = "*";
     private $interfaces = null;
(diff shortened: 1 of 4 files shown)


lib/Host.php
```diff
@@ -118,18 +120,17 @@ public function use($action): Host {
...
-        unset($options["SNI_server_certs"]);
-        $options["local_cert"] = $certificate;
-        if (isset($key)) {
-            $options["local_pk"] = $key;
+    public function encrypt($certificate): Host {
+        if (!$certificate instanceof ServerTlsContext) {
+            if (!$certificate instanceof Certificate) {
+                $certificate = new Certificate($certificate);
+            }
+            $certificate = (new ServerTlsContext)->withDefaultCertificate($certificate);
```
kelunik: Not sure about that, this still doesn't support SNI then.

kelunik:
@trowski This can be closed now, right?

kelunik:
I just verified that this is part of the `0.8.x` branch.

PR #213, opened 2018-01-05
0.8.x

bwoebi:
PR for discussion / review


lib/Cookie/MetaCookie.php
```diff
@@ -0,0 +1,220 @@
...
+     * @return string
+     */
+    protected function encodePath($path) {
+        return preg_replace_callback(
+            '/(?:[^A-Za-z0-9_\-\.~\/:%]+|%(?![A-Fa-f0-9]{2}))/',
+            function (array $matches) {
+                return rawurlencode($matches[0]);
+            },
+            $path
+        );
```
bwoebi: What's the point of that preg_replace_callback()? Why not just `rawurlencode($path)`? Same for Cookie class.

lib/Cookie/Cookie.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace Aerys\Cookie;
+
+class Cookie {
```
joelwurtz: This could be useful in amp/artax as well, maybe it would be nice to have an amp/http-message library in the future containing this code.

lib/Cookie/MetaCookie.php
```diff
@@ -0,0 +1,220 @@
+<?php
+
+namespace Aerys\Cookie;
+
+class MetaCookie extends Cookie {
```
joelwurtz: This could be useful in amp/artax as well, maybe it would be nice to have an amp/http-message library in the future containing this code.

lib/HttpStatus.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+namespace Aerys;
+
+/**
+ * @link https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml
+ */
+final class HttpStatus {
```
joelwurtz: This could be useful in amp/artax as well, maybe it would be nice to have an amp/http-message library in the future containing this code.

lib/Internal/Http2Driver.php
```diff
@@ -891,28 +871,35 @@ public function parser(Client $client, $settings = ""): \Generator {
...
-                $uri = $headerArray[":path"][0];
-                if (\strpos($uri, '?') !== false) {
-                    list($uri, $ireq->uriQuery) = \explode("?", $uri, 2);
+
+                if ($host !== null) {
+                    $uri = $scheme . "://" . \rawurldecode($host) . ":" . $port . $uri;
                 }
-                $ireq->uriPath = \rawurldecode($uri);
+
+                $ireq->uri = new Uri($uri);
```
bwoebi: Somewhat funny to construct an uri just to have the Uri class deconstruct it again. Possibly we should just have a dedicated static method creating an Uri instance given the actual values?

lib/Middleware/DeflateMiddleware.php
```diff
@@ -0,0 +1,165 @@
...
+     * @var int Minimum body length before body is compressed.
+     * @link http://webmasters.stackexchange.com/questions/31750/what-is-recommended-minimum-object-size-for-deflate-performance-benefits
+     */
+    private $minimumLength = 860;
+
+    /** @var string */
+    private $contentRegex = '#^(?:text/.*+|[^/]*+/xml|[^+]*\+xml|application/(?:json|(?:x-)?javascript))$#i';
+
+    /** @var int Minimum chunk size before being compressed. */
+    private $chunkSize = 8192;
```
bwoebi: What's now the source of truth for these options? Options class or here? Currently redundant, but never synchronized.

lib/Internal/Request.php
```diff
@@ -1,13 +1,13 @@
 <?php
 
-namespace Aerys;
+namespace Aerys\Internal;
```
bwoebi: This shouldn't be Internal. It's the raw request information, necessary for middlewares to mutate or access client.

lib/Internal/Vhost.php
```diff
@@ -29,37 +31,22 @@ class Vhost implements Monitor {
...
-        "ssl2"      => STREAM_CRYPTO_METHOD_SSLv2_SERVER,
-        "sslv2"     => STREAM_CRYPTO_METHOD_SSLv2_SERVER,
-        "ssl3"      => STREAM_CRYPTO_METHOD_SSLv3_SERVER,
-        "sslv3"     => STREAM_CRYPTO_METHOD_SSLv3_SERVER,
-        "any"       => STREAM_CRYPTO_METHOD_ANY_SERVER,
-    ];
-
     /** @Note Vhosts do not allow wildcards, only separate 0.0.0.0 and :: */
-    public function __construct(string $name, array $interfaces, callable $application, array $filters, array $monitors = [], HttpDriver $driver = null) {
+    public function __construct(string $name, array $interfaces, Responder $responder, array $middlewares = [], array $monitors = []) {
```
bwoebi: Oh. Why did you remove the possibility to specify an a specific/own http driver? (e.g. you want to be HTTP/2 only, or you want to implement a proxy http driver etc.)

lib/Middleware.php
```diff
@@ -2,6 +2,14 @@
...
+
 interface Middleware {
-    public function do(InternalRequest $ireq);
+    /**
+     * @param \Aerys\Request $request
+     * @param Responder $responder Request responder.
+     *
+     * @return \Amp\Promise<\Aerys\Response>
+     */
+    public function process(Request $request, Responder $responder): Promise;
```
bwoebi: wait … are we now really removing the access to internal structures here? :x (Request instead of Internal\Request)
I found that a rather powerful feature, when necessary. Not that _most_ use cases would need that, but it may be useful in some cases...

Also, that way you essentially removed the possibility to mutate the request before the responder can see it. Sure, I could fake a request with a my own Internal\Request instance, but, no. No.

lib/Internal/Request.php
kelunik: Middlewares will use `Aerys\Request`.

lib/Internal/Vhost.php
kelunik: Nobody ever used that. If you just want HTTP/2, that's better solved with an option. It's an internal detail that can change at any time if it's internal. We don't want to provide a BC guarantee here.

lib/Middleware.php
kelunik: No, you can still mutate the request before delegating to the responder, just like middlewares work in PSR-15. Which use cases do you have in mind?

lib/Response/EmptyResponse.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Aerys\Response;
+
+use Aerys\Response;
+use Amp\ByteStream\InMemoryStream;
+
+class EmptyResponse extends Response {
+    public function __construct(array $headers = [], int $code = 200, string $reason = null) {
```
rdlowrey: Default status code for an empty response should probably be `204 No Content` instead of `200`

lib/Middleware.php
bwoebi: Well, then please show me how (while still fulfilling the interface contracts).

kelunik: Request is supposed to be mutable, but isn't yet. I guess that's on @trowski's TODO list. We discussed that earlier.

lib/Root.php
```diff
@@ -292,66 +299,62 @@ private function coalesceIndexPath(string $dirPath): \Generator {
...
 
-    private function respond($fileInfo, Request $request, Response $response) {
+    private function respondFromFileInfo($fileInfo, Request $request): \Generator {
         // If the file doesn't exist don't bother to do anything else so the
         // HTTP server can send a 404 and/or allow handlers further down the chain
         // a chance to respond.
         if (empty($fileInfo->exists)) {
-            return;
+            $status = HttpStatus::NOT_FOUND;
+            return new Response\HtmlResponse(makeGenericBody($status), [], $status);
```
rdlowrey: Using `HtmlResponse` here defeats the Server's default [content type and charset options](https://github.com/amphp/aerys/blob/0.8.x/lib/Options.php#L18-L19)

trowski: That was being defeated in several areas. This is an issue we should probably look at, perhaps with an error response handler.

lib/Response/EmptyResponse.php
trowski: Yep, oversight.

lib/Middleware.php
trowski: @bwoebi I planned on adding options to mutate `Aerys\Request`, exposing only the things that should be mutated.

lib/Internal/Request.php
trowski: `Aerys\Request` will have a method that will return client information.

lib/Root.php
kelunik: A server default shouldn't be a thing. Changing the server default breaks everything assuming that default otherwise.

lib/Middleware/DeflateMiddleware.php
trowski: Updating the `Options` class to remove unused settings was on my TODO list. This class obviously needs settings for these options yet.

lib/Internal/Http2Driver.php
trowski: Yes, I think the Uri class could use some improvement in this regard, as the only way to construct one is to have a valid uri string pre-built.

kelunik: Which is totally fine. The `Uri` class would have to validate the parts either way.

test/WatcherProcessTest.php
```diff
@@ -47,8 +47,8 @@ public function __get($prop) {
                     }
                     throw new \Error("Invalid property access");
                 }
-                public function stop(): \Generator {
-                    return parent::doStop();
+                public function stop(): Promise {
+                    return new Coroutine(parent::doStop());
```
kelunik: This should just return instead of New Coroutine

bwoebi:
Is it possible to kill Host off too? If you drop the vhosts, and have the config file return a mere responder?

At least Host feels a bit wrong as it indicates still some vhost management to me.

E.g. you still expose Host::name() which is totally pointless without vhosts.
The interface to expose / TLS settings can be put into the Options object instead. (as they are now global).

bin/aerys
```diff
@@ -111,12 +105,17 @@ $climate->arguments->add([
...
+
 try {
     if ($console->isArgDefined("help")) {
         echo $help;
         exit(0);
     }
-} catch (Exception $e) {
+
+    if (!$console->isArgDefined("config")) {
+        throw new \Error("A config file must be defined using -c or --config");
```
kelunik: Why this manual check?

trowski: Running `bin/aerys --help` would say the command arguments were invalid. To fix that I had to remove the requirement for config, but then manually check for it if help wasn't requested.

lib/Response.php
```diff
@@ -57,6 +61,18 @@ public function __construct(
         }
     }
 
+    public function __destruct() {
```
kelunik: I think this should be final, so child classes can't remove it. If they need their own destruct, they can just register a handler.

lib/Options.php
```diff
@@ -2,259 +2,378 @@
...
-    public function __get(string $property) {
-        if (\property_exists($this, $property)) {
-            return $this->{$property};
-        }
-        // Use \Amp\Struct::generateStructPropertyError() to get a nice message
-        // with a possible suggestion for the correct name
-        throw new \Error(
-            $this->generateStructPropertyError($property)
-        );
+    public function inDebug(): bool {
```
kelunik: `isDebug`

lib/HttpStatus.php
kelunik: Moved to `amphp/http` now.

lib/Message.php
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * @param \Amp\ByteStream\InputStream|string|null $stringOrStream
+     * @param string[][] $headers
+     * @param int $code Status code.
+     * @param string|null $reason Status code reason.
+     *
+     * @throws \Error If one of the arguments is invalid.
+     */
+    public function __construct(array $headers = []) {
```
kelunik: I don't think this constructor makes sense to have.

lib/Message.php
```diff
@@ -0,0 +1,155 @@
...
+     * @param string[] $values
+     *
+     * @return bool
+     *
+     * @throws \Error If the given value cannot be converted to a string and is not an array of values that can be
+     *     converted to strings.
+     */
+    private function isValueValid(array $values): bool {
+        foreach ($values as $value) {
+            if (\preg_match("/[^\t\r\n\x20-\x7e\x80-\xfe]|\r\n/", $value)) {
```
kelunik: Better use single quotes for every regex that includes `\xXX`.

lib/Http1Driver.php
```diff
@@ -223,13 +223,14 @@ public function parser(): \Generator {
                 throw new ClientException("Bad Request: multiple host headers", Status::BAD_REQUEST);
             }
 
-            if (!\preg_match("#^(?:https?://)?([^\s:]+)(?::(\d+))?$#i", $headers["host"][0], $matches)) {
+            if (!\preg_match("#^([A-Z\d\.\-]+|\[[\d:]+\])(?::([1-9]\d*))?$#i", $headers["host"][0], $matches)) {
```
kelunik: What about internationalized domain names? Do they have to be ASCII encoded in `host` headers?

trowski: Yes, host names must be ASCII in host headers.

joshdifabio:
So you can _only_ attach objects to a request, and you cannot attach multiple objects of the same class?

Let's say I want to attach a timestamp to a request; I would have to create a new class, e.g. `RequestTimeAttribute`, which would simply hold a reference to my `DateTimeImmutable` object? (If I don't create a new class, some other piece of code might silently override my value with a different instance of `DateTimeImmutable` which has some other meaning.)

Let's say I want to attach a string attribute to the request; I would have to create a class for this?

What is the purpose of this design? Is the idea that attributes are always namespaced or is it about providing type safety/IDE hints when calling `Request::get()`?

kelunik:
@joshdifabio Both. :-)

And yes, that's the downside.

joshdifabio:
Okay. I've just seen #216 so I'll comment there instead.

lib/Client.php
```diff
@@ -521,23 +523,24 @@ private function write(string $data, bool $close = false): Promise {
...
      *
      * @throws \Aerys\ClientException If the client has disconnected.
      */
-    private function send(string $data): string {
+    private function send(string $data, bool $strict = false): string {
         $bytesWritten = @\fwrite($this->socket, $data);
-        if ($bytesWritten === false
-            || ($bytesWritten === 0 && (!\is_resource($this->socket) || @\feof($this->socket)))
-        ) {
+        if ($bytesWritten === false || ($strict && $bytesWritten === 0)) {
```
kelunik: What if the write buffer is full? Then it will just close the connection?

lib/Internal/HPack.php
```diff
@@ -75,43 +75,67 @@ class HPack {
...
+        self::$huffmanLookup = self::huffmanLookupInit();
+        self::$huffmanCodes = self::huffmanCodesInit();
+        self::$huffmanLengths = self::huffmanLengthsInit();
+
+        foreach (\array_column(self::TABLE, 0) as $index => $name) {
+            if (isset(self::$indexMap[$name])) {
+                continue;
+            }
+
+            self::$indexMap[$name] = $index + 1;
```
kelunik: Why wasn't this present before?

lib/Client.php
bwoebi: Yep, writing this style of code is only fine in an onWritable callback as there we're guaranteed free space in the write buffer.

lib/Internal/HPack.php
bwoebi: Looks like I missed that one small optimization ;-)

trowski: Technically the map should be to an array of potential header values, as some of the headers are in the table twice. As a server we really only stand to benefit from the predefined values for `:status`, which are tiny anyway, so there's not a lot to be gained, but something that could be improved upon further.

lib/Client.php
trowski: `$strict` is only set to true in the call from the `onWritable` callback.

bwoebi: Then I recommend calling this `$fromOnWritable` to explicitly hint that it needs to be called from within onWritable

PR #231, opened 2018-03-13
Prevent empty header arrays

### LINKED ISSUE #230: Prevent empty header arrays
`Message::setHeader()` shouldn't create an array entry if the header array is empty, but instead delete the current entry if it exists. Similar for `Message::addHeader()`. There should also be a `array_values()` call to ensure no string keys are present in the array.

ekinhbayar:
Fixes #230

### DIFF
--- a/src/Message.php
+++ b/src/Message.php
@@ -63,7 +63,12 @@ abstract class Message {
         \assert($this->isNameValid($name), "Invalid header name");
 
         if (\is_array($value)) {
-            $value = \array_map("strval", $value);
+            if (!$value) {
+                $this->removeHeader($name);
+                return;
+            }
+
+            $value = \array_values(\array_map("strval", $value));
         } else {
             $value = [(string) $value];
         }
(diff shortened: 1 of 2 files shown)


kelunik:
@ekinhbayar Do you want to add a test for the interaction of `setHeader` and `hasHeader`?

kelunik:
Thanks!

src/Message.php
```diff
@@ -86,7 +91,12 @@ public function addHeader(string $name, $value) {
         \assert($this->isNameValid($name), "Invalid header name");
 
         if (\is_array($value)) {
-            $value = \array_map("strval", $value);
+            if (!$value) {
+                $this->removeHeader($name);
```
trowski: The header should not be removed from this method, this should just `return`.

kelunik: Oops, yes. Missed that.

kelunik: Fixed via acb8019923f4e8fed6abcff1fdf2dae9849f5c54.

Issue #242, opened 2018-03-17
Request documentation


kelunik:
Assigned to @krlv.

PR #245, opened 2018-03-19
Request documentation update

krlv:
PR for #242:

1. Updated Jekyll config (`gems` -> `plugins`)
2. Updated `docs/README.md`: `git submodule` explanation added
3. Removed links to missing classes in `docs/classes/index.md`
4. Added stubs for new `docs/classes/message.md`, `docs/classes/requestbody.md` and `docs/classes/requesthandler.md`
5. Updated `docs/classes/request.md`
6. Changed the order of setter and getter of HTTP method (to keep it consistent)

### DIFF
--- a/docs/classes/index.md
+++ b/docs/classes/index.md
@@ -2,24 +2,16 @@
 title: Classes
 permalink: /classes/
 ---
-Aerys provides a set of classes and interfaces, as well as [functions](functions.md):
+The `http-server` package provides a set of classes and interfaces, as well as [functions](functions.md):
 
-- [`BodyParser`](bodyparser.md) &mdash; Parser for bodies
-- [`Bootable`](bootable.md) &mdash; Registers entry point for [`Server`](server.md) and [`Logger`](logger.md)
 - [`Client`](client.md) &mdash; Client connection related information
-- [`CommandClient`](commandclient.md) &mdash; Controls the server master process
-- [`FieldBody`](fieldbody.md) &mdash; Field body message container (via [`BodyParser`](bodyparser.md))
-- [`Filter`](filter.md) &mdash; Defines a filter callable in `do()` method
-- [`Host`](host.md) &mdash; Registers a virtual host
 - [`HttpDriver`](httpdriver.md) &mdash; Driver for interaction with the raw socket
-- [`InternalRequest`](internalrequest.md) &mdash; Request related information
 - [`Logger`](logger.md) &mdash; PSR-3 compatible logger
+- [`Message`](message.md) &mdash; An abstract class represents HTTP message
 - [`Options`](options.md) &mdash; Accessor of options
-- [`ParsedBody`](parsedbody.md) &mdash; Holds request body data in parsed form
-- [`Request`](request.md) &mdash; General request interface for responder callables
+- [`Request`](request.md) &mdash; Request class for request handler callables
+- [`RequestBody`](request-body.md) &mdash; Request body
+- [`RequestHandler`](request-handler.md) &mdash; Request handler interface 
 - [`Response`](response.md) &mdash; General response interface for responder callables
-- [`Router`](router.md) &mdash; Manages and accumulates routes
 - [`Server`](server.md) &mdash; The server, tying everything together
 - [`ServerObserver`](serverobserver.md) &mdash; Registers method to be notified upon Server state changes
-- [`Websocket`](websocket.md) &mdash; General websocket connection manager
-- [`Websocket\Endpoint`](websocket-endpoint.md) &mdash; Provides API to communicate with a websocket client
--- a/docs/classes/request.md
+++ b/docs/classes/request.md
@@ -2,86 +2,126 @@
 title: Request
 permalink: /classes/request
 ---
+# Request
+
+The **`Request`** class represents an HTTP request. It's used in [request handlers](request-handler.md) and [`middleware`](middleware.md).
 
 * Table of Contents
 {:toc}
 
-The `Request` interface generally finds its only use in responder callables (or [`Websocket::onOpen()`](websocket.html#onopenint-clientid-handshakedata)). [`Middleware`s](middleware.md) do never see the `Request`; the `StandardRequest` class is supposed to be a simple request API reading from and manipulating [`InternalRequest`](internalrequest.md) under the hood.
+## Constructor
+
+```php
+public function __construct(
+    Client $client, 
+    string $method, 
+    Psr\Http\Message\UriInterface $uri, 
+    string[] | string[][] $headers = [], 
+    RequestBody | Amp\ByteStream\InputStream | string | null $body = null, 
+    string $protocol = "1.1"
+)
+```
+
+### Parameters
+
+|[`Client`](client.md)|`$client`|The client sending the request|
+|`string`|`$method`|HTTP request method|
+|[`Psr\Http\Message\UriInterface`](https://www.php-fig.org/psr/psr-7/#35-psrhttpmessageuriinterface)|`$uri`|The full URI being requested, including host, port, and protocol|
+|`string[]`<br />`string[][]`|`$headers`|An array of strings or an array of string arrays.|
+|[`RequestBody`](request-body.md)<br />[`Amp\ByteStream\InputStream`](https://amphp.org/byte-stream/)<br />`string`<br />`null`|`$body`|Request body|
+|`string`|`$protocol`|HTTP protocol version|
+
+## `getClient(): Client`
+
+Returns the [`Сlient`](client.md) sending the request
 
 ## `getMethod(): string`
 
-Returns the used method, e.g. `"GET"`.
+Returns the HTTP method used to make this request, e.g. `"GET"`.
 
-## `getUri(): string`
+## `setMethod(string $method)`
 
-Returns the requested URI (path and query string), e.g. `"/foo?bar"`.
+Sets the request HTTP method.
+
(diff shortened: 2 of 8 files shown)


docs/classes/index.md
```diff
@@ -2,24 +2,16 @@
 title: Classes
 permalink: /classes/
 ---
-Aerys provides a set of classes and interfaces, as well as [functions](functions.md):
+`http-server` package provides a set of classes and interfaces, as well as [functions](functions.md):
```
kelunik: The `http-server` package...

docs/classes/request.md
```diff
@@ -2,86 +2,128 @@
 title: Request
 permalink: /classes/request
 ---
+# Request
+
+Class **`Request`** extends [`Message`](message.md)
```
kelunik: This should only be an internal detail IMO, maybe we should make `Message` a trait?

docs/classes/request.md
```diff
@@ -2,86 +2,128 @@
...
+
+The `Request` class represents an HTTP request. It's used in [request handlers](requesthandler.md) and [`middleware`](middleware.md).
```
kelunik: `requesthandler.md` should be `request-handler.md`.

docs/classes/request.md
```diff
@@ -2,86 +2,128 @@
...
 
 * Table of Contents
 {:toc}
 
-The `Request` interface generally finds its only use in responder callables (or [`Websocket::onOpen()`](websocket.html#onopenint-clientid-handshakedata)). [`Middleware`s](middleware.md) do never see the `Request`; the `StandardRequest` class is supposed to be a simple request API reading from and manipulating [`InternalRequest`](internalrequest.md) under the hood.
+## `Constructor`
```
kelunik: `Constructor` shouldn't be in code tags IMO.

docs/classes/request.md
```diff
@@ -2,86 +2,128 @@
...
+|[`Client`](client.md)|`$client`|The client sending the request|
+|`string`|`$method`|HTTP request method|
+|[`Psr\Http\Message\UriInterface`](https://www.php-fig.org/psr/psr-7/#35-psrhttpmessageuriinterface)|`$uri`|The full URI being requested, including host, port, and protocol|
+|`string[]`<br />`string[][]`|`$headers`|An array of strings or an array of string arrays.|
+|[`RequestBody`](requestbody.md)<br />[`Amp\ByteStream\InputStream`](https://amphp.org/byte-stream/)<br />`string`<br />`null`|`$body`|Request body|
+|`string`|`$protocol`|HTTP protocol version|
+
+## `getClient(): Client`
+
+Returns the [`client`](client.md) sending the request
```
kelunik: Should be `Client` (uppercase)

docs/classes/request.md
```diff
@@ -2,86 +2,128 @@
...
+Sets a new protocol version number for the request.
 
-## `getHeaderArray(string): array<string>`
+## `setHeader(string $name, string | string[] $value)`
 
-Gets an array with headers. HTTP allows for multiple headers with the same name, so this returns an array. Usually only a single header is needed and expected, in this case there is `getHeader()`.
+Sets the named header to the given value.
 
-## `getAllHeaders(): array<array<string>>`
+If header `$name` is not valid or `$value` is invalid, [`\Error`](http://php.net/manual/en/class.error.php) is thrown.
```
kelunik: This only happens when assertions are enabled, currently.

docs/classes/request.md
```diff
@@ -2,86 +2,128 @@
...
 
-Gets an array with headers. HTTP allows for multiple headers with the same name, so this returns an array. Usually only a single header is needed and expected, in this case there is `getHeader()`.
+Sets the named header to the given value.
 
-## `getAllHeaders(): array<array<string>>`
+If header `$name` is not valid or `$value` is invalid, [`\Error`](http://php.net/manual/en/class.error.php) is thrown. 
 
-Returns all the headers in an associative map with the keys being normalized header names in lowercase.
+{:.note}
+> To verify header `$name` and `$value` [`Message::setHeader`](message.md#setheaderstring-name-string--string-value) is used
```
kelunik: I don't think that's relevant to a reader here.

docs/classes/request.md
```diff
@@ -2,86 +2,128 @@
...
 
-The parameter `$bodySize` defaults to `-1` to take the globally configured maximum, otherwise it's the maximum body size.
+## `setBody(RequestBody | InputStream | string | null $stringOrStream)`
 
-There also exists a [`parseBody()`](parsedbody.md) function for processing of a typical HTTP form data.
+Sets the stream for the message body
 
-## `getCookie(string): string | null`
+{:.note}
+> Using a string will automatically set the `Content-Length` header to the length of the given string. Using an [`InputStream`](https://amphp.org/byte-stream/) or [`Body`](requestbody.md) instance will remove the `Content-Length` header.
```
kelunik: The `content-length` header isn't changed in case a stream is passed.

docs/classes/request.md
```diff
@@ -2,86 +2,128 @@
...
+## `getCookie(string): RequestCookie | null`
 
-## `getLocalVar(string)` / `setLocalVar(string, $value)`
+Gets a [cookie](https://amphp.org/http/cookies) value by name or `null`.
 
-These methods are only important when using [`Middleware`s](middleware.md). They manipulate the [`InternalRequest->locals`](internalrequest.html#locals) array.
+## `setCookie(RequestCookie $cookie)`
 
-## `getOption(string)`
+Adds a [`$cookie`](https://amphp.org/http/cookies) to the response.
```
kelunik: Should be `Cookie` instead of `$cookie` IMO. Also, needs to be `request` not `response`.

docs/classes/requestbody.md
```diff
@@ -0,0 +1,10 @@
+---
+title: RequestBody
+permalink: /classes/requestbody
```
kelunik: Should be `request-body`.

docs/classes/request.md
krlv: Removed.

docs/classes/request.md
krlv: Renamed

docs/classes/request.md
krlv: Fixed.

docs/classes/request.md
krlv: Fixed.

docs/classes/request.md
krlv: Should I leave info block about it (with details about `assert.exception` settings) or it's better to remove it at all?

docs/classes/request.md
krlv: Actually, it's about `setHeader`: https://github.com/amphp/http-server/pull/245/files#diff-f9828b2f68601cc1d949f899eee34469R62

docs/classes/requestbody.md
krlv: Renamed.

docs/classes/request.md
krlv: Fixed. Also fixed mistake in the code doc block

docs/classes/request.md
krlv: To be really honest, I copy-pasted this part from phpdoc of the `setBody` method. 

Should I change the comment or remove `content-length` header for cases when a stream passed to the func?

docs/classes/request.md
kelunik: Not sure, generally `Error` instances shouldn't be caught, anyway. @trowski Thoughts?

docs/classes/request.md
kelunik: I'm not entirely sure. @trowski Thoughts?

trowski: The docblock is wrong, it must have not been updated when that behavior was changed. `content-length` is only changed if a string or null is passed. Please change both.

docs/classes/request.md
trowski: I'd just leave it out of the docs for now as @kelunik said `Error` shouldn't be caught anyway. Both incoming and outgoing header blocks are validated regardless of the assertion setting, we only have the assertions there so the error is thrown in a relevant location during development.

trowski:
Thanks @krlv, we really appreciate your help updating the docs!

kelunik:
Thanks!

kelunik:
@krlv Just tried it out locally, `Deprecation: The 'plugins' configuration option has been renamed to 'plugins_dir'. Please update your config file accordingly.` now, all these useless renames, lol.

krlv:
@kelunik that's strange 🤔 Here is my output of `bundle exec jekyll serve` command for Jekyll 3.6.2:

```bash
Configuration file: /Users/eugene/Projects/http-server/docs/_config.yml
            Source: /Users/eugene/Projects/http-server/docs
       Destination: /Users/eugene/Projects/http-server/docs/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
                    done in 1.319 seconds.
        ** ERROR: directory is already being watched! **

        Directory: /Users/eugene/Projects/http-server/docs/.shared/asset

        is already being watched through: /Users/eugene/Projects/http-server/docs/.shared/asset

        MORE INFO: https://github.com/guard/listen/wiki/Duplicate-directory-errors
 Auto-regeneration: enabled for '/Users/eugene/Projects/http-server/docs'
    Server address: http://127.0.0.1:4000/http-server/
```

krlv:
Oh, looks like I'm not using the latest version which is v7.3.7

PR #246, opened 2018-03-20
Update classes/request docs

krlv:
Small doc updates that were requested in PR #245

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -182,7 +182,7 @@ class Request extends Message {
 
     /**
      * Sets the stream for the message body. Note that using a string will automatically set the Content-Length header
-     * to the length of the given string. Using an InputStream or Body instance will remove the Content-Length header.
+     * to the length of the given string.
      *
      * @param RequestBody|InputStream|string|null $stringOrStream
      *
--- a/src/Request.php
+++ b/src/Request.php
@@ -51,12 +51,12 @@ class Request extends Message {
         $this->uri = $uri;
         $this->protocol = $protocol;
 
-        if (!empty($headers)) {
-            $this->setHeaders($headers);
-        }
-
         if ($body !== null) {
             $this->setBody($body);
+        }
+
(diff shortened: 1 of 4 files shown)


src/Request.php
```diff
@@ -182,7 +182,7 @@ public function getBody(): RequestBody {
 
     /**
      * Sets the stream for the message body. Note that using a string will automatically set the Content-Length header
-     * to the length of the given string. Using an InputStream or Body instance will remove the Content-Length header.
```
kelunik: I guess we should have a note here, that the `content-length` header needs to be removed manually if you change the body stream size.

Maybe we should remove it automatically instead? If you want to set it manually, you can set it _after_ calling `setBody()`, which is an auto-header opt-in instead of an opt-out where users have to care and easily make mistakes.

@trowski Thoughts?

trowski: It is probably better to remove the `content-length` header if a stream is used. The order of setting headers and the body needs to be flipped in the constructor though, as the header shouldn't be removed if it was given to the constructor.

krlv: 1. Flipped the order of setting headers and the body
2. Removed `Content-Length` in `setBody` for cases when instance of `RequestBody` or `InputStream` passed to the method
3. Updated tests

Please, review it again @trowski @kelunik

kelunik: @krlv The text in the line this comment is for should be restored then.

kelunik: @krlv Could you update `Response` as well, so it's consistent?

krlv: @kelunik fixed

kelunik: @krlv Thanks, there's an additional space here now.

krlv: @kelunik fixed. I'm very sorry that it required so many fixes for such a small amount of code (not even a code, but docs! 😞 )

kelunik:
Thanks!

Issue #226, opened 2018-03-12
Documentation update


kelunik:
The 0.8.x will change a lot of things, so we need a big documentation update. How should we organize this so we don't end up doing double work and inconsistencies?

krlv:
Maybe you can split it into separate issues (eg. update `middleware` doc; update `classes/client` doc, etc.) and then assign each issue to a contributor (or to yourself if no one volunteers 😞). 

Also, it would be great if you can give a hint what should be added/updated in each case.

krlv:
Btw, I can try to start with `docs/classes/request.md` update, if no one is working on it.

kelunik:
@krlv Just started yesterday with the documentation update, don't have a fixed TODO list, yet. I opened https://github.com/amphp/http-server/issues/242 and "assigned" it to you, assignment doesn't work directly to external contributors. :-(

krlv:
I guess external contributors should leave a comment in the issue they want to work on. Saw this couple of times during Hacktoberfest.

krlv:
@kelunik I can continue with `docs/classes/message.md` docs (if you not working on ti at the moment)

kelunik:
@krlv Better skip that for now, because it's likely to be renamed or moved into `amphp/http`. You could work on `Response` if you want.

krlv:
Ok, will do `Response` update then.

krlv:
@kelunik let me know if I can help with something else.

kelunik:
@krlv Thanks for the offer! You could look into documenting `Options` if you want. If you have even more time, you could also update the options on https://amphp.org/http-server/production#options and remove the no longer existing ones there.

kelunik:
I think only the production options are missing now. :-)

kelunik:
Closing, as we're complete.

trowski:
Server still needs an update.

PR #250, opened 2018-03-22
Update Options docs

krlv:
#226 Documentation update for `Options`

### DIFF
--- a/docs/classes/options.md
+++ b/docs/classes/options.md
@@ -3,180 +3,225 @@ title: Options
 permalink: /classes/options
 ---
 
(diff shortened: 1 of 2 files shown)


krlv:
prnscr_warning_info
@kelunik I used `note` blocks to show default values and `warning` blocks for validation rules. Does it make easier to read the docs this way, wdyt?

kelunik:
Thanks! I'll keep the notes and warnings for now, but I think we'll have to work out a better concept in the future, as the boxes take too much attention in this case compared to the actual description of the option.

PR #254, opened 2018-03-26
Convert convertible types to string on setBody()

kelunik:
Two separate methods would really have been cleaner, as that would have allowed people to use strict types if they prefer.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -220,17 +220,23 @@ final class Request extends Internal\Message {
             return;
         }
 
-        if ($stringOrStream !== null && !\is_string($stringOrStream)) {
+        try {
+            // Use method with string type declaration, so we don't need to implement our own check.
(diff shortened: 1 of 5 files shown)


trowski:
Perhaps, but then the constructor probably would still have been mixed, so then we'd need type-checking logic there.

This method matches better with how we handle mixed types for header values, so :+1:.

trowski:
Giving this more thought, I assume this is largely to accommodate objects with a `__toString()` method, which perhaps we should explicitly check for rather than relying on casting here. Passing a number or bool for the body is certainly a coding error, whereas for headers it is not.

kelunik:
I've run into this because I passed an integer as body, not because of toString. It was a simple visitor counter.

kelunik:
@trowski Should we merge this or not?

PR #257, opened 2018-04-01
Server class docs

krlv:
Missing piece for #226

### DIFF
--- a/docs/classes/server.md
+++ b/docs/classes/server.md
@@ -3,71 +3,77 @@ title: Server
 permalink: /classes/server
 ---
 
-{:.warning}
-> This page needs an update.
-
 * Table of Contents
 {:toc}
 
 The `Server` instance controls the whole listening and dispatches the parsed requests.
 
-## `attach(ServerObserver)`
+The `Server` class has a [custom debug output](http://php.net/manual/en/language.oop5.magic.php#object.debuginfo), which isn't part of the public API and which shouldn't be relied upon.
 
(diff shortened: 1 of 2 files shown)


docs/classes/server.md
```diff
@@ -3,71 +3,103 @@ title: Server
...
+
+Starts the server.
+
+## `stop(int $timeout = 3000ms): Promise`
+
+Stops the server. `$timeout` is the number of milliseconds to allow clients to gracefully shutdown before forcefully closing.
+
+The returned `Promise` will resolve when the server has successfully been stopped.
+
+## `__debugInfo(): array`
```
kelunik: IMO `__debugInfo()` shouldn't be documented or explicitly documented as not being public API.

krlv:
@kelunik Moved info about debug output to the main description. Wdyt, should I remove it completely?

docs/classes/server.md
```diff
@@ -3,71 +3,99 @@ title: Server
...
-{:.warning}
-> This page needs an update.
-
 * Table of Contents
 {:toc}
 
 The `Server` instance controls the whole listening and dispatches the parsed requests.
 
-## `attach(ServerObserver)`
+The `Server` class has [custom debug output](http://php.net/manual/en/language.oop5.magic.php#object.debuginfo):
```
kelunik: I wouldn't remove it completely, but remove the output and add: custom debug output, which isn't part of the public API and which shouldn't be relied upon.

krlv: Fixed.

kelunik:
Thanks!

PR #260, opened 2018-06-08
Fix example in readme

vlakarados:
Additional required argument logger was missing. Added NullLogger instance for zero configuration.


kelunik:
Thanks!

PR #267, opened 2018-09-18
Log error if zlib is disabled

Zvax:
Using compression defaults to true in Options which can lead to an `Undefined constant ZLIB_ENCODING_GZIP` error on a php built without zlib.

This check prevents the compression middleware to be added to the stack on these systems while advising user to install zlib.


kelunik:
Thanks!

Zvax:
My pleasure, cheers!

PR #272, opened 2019-02-13
Fix middleware example

gquemener:


docs/classes/middleware.md
```diff
@@ -36,12 +36,12 @@ $handler = new CallableRequestHandler(function (Request $request) {
 
 $middleware = new class implements Middleware {
     public function handleRequest(Request $request, RequestHandler $next): Promise {
-        return call(function () {
+        return call(function () use ($request, $next) {
```
rybakit:
```suggestion
        return call(static function () use ($request, $next) {
```

gquemener: This modification would not be consistent with all the other php examples in this repository.

Thus, I suggest performing it (across all php examples) in a separate PR.

WDYT?

kelunik: I don't care, personally I'd do it right away in this.

PR #274, opened 2019-05-06
Fix isWaitingOnResponse and getPendingResponseCount

### LINKED ISSUE #273: how can i disable connection timeout?
im trying to create a [sse](https://en.wikipedia.org/wiki/Server-sent_events) server. so i need keep connection alive. is there a way for disable timeout?

kelunik:
Responses are now pending until they're fully written.

This might technically be a BC break.

Fixes #273.

### DIFF
--- a/src/Driver/RemoteClient.php
+++ b/src/Driver/RemoteClient.php
@@ -96,7 +96,7 @@ final class RemoteClient implements Client
     /** @var int */
     private $pendingResponses = 0;
 
-    /** @var bool  */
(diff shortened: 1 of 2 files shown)


kelunik:
This isn't really correct, will have another look.

trowski:
Pushed another commit that might fix slow incoming requests. If it doesn't, feel free to force-push over it.

trowski:
Merged manually.

PR #275, opened 2019-05-30
Fix PHPStorm meta

asdfpowdfeuavxz:
i use this


kelunik:
According to the [official example](https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata#PhpStormAdvancedMetadata-Deprecated:Legacymetadataformat(2016.1andearlier)) the current way is the correct format. Maybe we should move away from the legacy format and to the new format?

kelunik:
There hasn't been any response, so I'm closing this due to inactivity.

PR #277, opened 2019-06-17
Update Response.php

bosunski:
Changed `The response body must a string, null, or instance of ...` to The response body must be a string, null, or instance of ...`


kelunik:
Thanks!

PR #281, opened 2019-07-23
Remove redundant check

assertchris:
`$clientFactory` will never have a value. Perhaps it used to be a constructor argument, or this line was copied from elsewhere.

### DIFF
--- a/src/Server.php
+++ b/src/Server.php
@@ -113,7 +113,7 @@ final class Server
 
         $this->logger = $logger;
         $this->options = $options ?? new Options;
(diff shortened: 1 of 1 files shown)


kelunik:
Rather, it was never added as constructor argument, I guess.

kelunik:
Thanks!

assertchris:
👍

trowski:
I must have forgotten to add the constructor argument or it got lost in a merge… maybe I'll add it later.

PR #284, opened 2019-09-29
Improve Trailers API

kelunik:


src/Trailers.php
```diff
@@ -95,7 +94,7 @@ public function getFields(): array
     /**
      * @return Promise<Message>
      */
-    public function awaitMessage(): Promise
+    public function await(): Promise
```
trowski: This method is used in several tests.

kelunik: Fixed, thanks!

PR #286, opened 2019-09-30
Add option to limit control frames between data frames

trowski:


src/Driver/Http2Driver.php
```diff
@@ -743,6 +730,41 @@ private function parser(string $settings = null): \Generator
...
+                    $lastReset = $now;
+                    $frameCount = 0;
+                    $bytesReceived = 0;
+                }
+
+                switch ($type) {
+                    case self::DATA:
+                    case self::HEADERS:
+                    case self::CONTINUATION:
+                        $controlFrameCount = 0;
```
kelunik: I don't think these should reset the count to `0`. Maybe take some inspiration from [nginx](https://github.com/nginx/nginx/commit/af0e284b967d0ecff1abcdce6558ed4635e3e757)?

trowski: I prefer that approach as well. I'll put together another PR.

trowski:
Closing in favor of #287.

PR #287, opened 2019-10-02
Refactor flood prevention

trowski:
Dropping arbitrary average frame size in favor of nginx-inspired flood protection based on the number of payload (header or data) frames received vs. the overall number of bytes received, including frame headers. This protects against both control frame floods and tiny data frame floods.

### DIFF
--- a/src/Driver/Http2Driver.php
+++ b/src/Driver/Http2Driver.php
@@ -31,7 +31,6 @@ final class Http2Driver implements HttpDriver
     public const DEFAULT_MAX_FRAME_SIZE = 1 << 14;
     public const DEFAULT_WINDOW_SIZE = (1 << 16) - 1;
 
-    private const SMALL_WINDOW_UPDATE_THRESHOLD = 1024;
     private const MINIMUM_WINDOW = (1 << 15) - 1;
     private const MAX_INCREMENT = (1 << 16) - 1;
 
@@ -626,7 +625,6 @@ final class Http2Driver implements HttpDriver
 
         $totalBytesReceivedSinceReset = 0;
         $payloadBytesReceivedSinceReset = 0;
-        $smallWindowUpdates = 0;
         $lastReset = $lastStreamOpening = $this->timeReference->getCurrentTime();
         $continuation = false;
 
@@ -1288,14 +1286,6 @@ final class Http2Driver implements HttpDriver
 
                             $windowSize = \unpack("N", $buffer)[1];
                             $buffer = \substr($buffer, 4);
-
-                            if ($windowSize < self::SMALL_WINDOW_UPDATE_THRESHOLD && ++$smallWindowUpdates > 10) {
-                                throw new Http2ConnectionException(
-                                    $this->client,
-                                    "Too many small window updates",
-                                    self::ENHANCE_YOUR_CALM
-                                );
-                            }
 
                             if ($id) {
                                 if (!isset($this->streams[$id])) {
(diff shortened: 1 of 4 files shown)


src/Driver/Http2Driver.php
```diff
@@ -706,27 +707,37 @@ private function parser(string $settings = null): \Generator
...
 
             while (true) {
-                if (++$frameCount === $maxFramesPerSecond) {
-                    $now = $this->timeReference->getCurrentTime();
-                    if ($lastReset === $now && $bytesReceived / $frameCount < $minAverageFrameSize) {
+                $now = $this->timeReference->getCurrentTime();
+                if ($lastReset === $now) {
+                    // Inspired by nginx flood detection:
+                    // https://github.com/nginx/nginx/commit/af0e284b967d0ecff1abcdce6558ed4635e3e757
+                    if ($totalBytesReceivedSinceReset > $payloadBytesReceivedSinceReset + 8192) {
```
kelunik: A large header might exceed this, no?

src/Driver/Http2Driver.php
```diff
@@ -706,27 +707,37 @@ private function parser(string $settings = null): \Generator
...
+                    $totalBytesReceivedSinceReset = 0;
+                    $payloadBytesReceivedSinceReset = 0;
+                }
+
+                if (empty($this->streams) && $lastStreamOpening < $now - $timeout) {
+                    throw new Http2ConnectionException(
+                        $this->client,
+                        "Too much time elapsed with non-payload frames",
+                        self::ENHANCE_YOUR_CALM
+                    );
```
kelunik: Does this send the unprocessed stream ID to the client?

src/Driver/Http2Driver.php
```diff
@@ -1270,6 +1287,14 @@ private function parser(string $settings = null): \Generator
                             $windowSize = \unpack("N", $buffer)[1];
                             $buffer = \substr($buffer, 4);
 
+                            if ($windowSize < self::SMALL_WINDOW_UPDATE_THRESHOLD && ++$smallWindowUpdates > 10) {
```
kelunik: `$smallWindowUpdates` should be reset somewhere, otherwise it will definitely happen on a long living connection.

src/Driver/Http2Driver.php
```diff
@@ -1633,7 +1660,8 @@ function (int $bodySize) use ($id) {
                 }
             }
         } catch (Http2ConnectionException $exception) {
-            yield $this->shutdown(null, $exception);
+            $this->shutdown(null, $exception);
```
kelunik: Why is this no longer awaited?

src/Driver/Http2Driver.php
trowski: Yes, headers and continuation frames are considered payload data.

src/Driver/Http2Driver.php
trowski: Yes, the shutdown handler will send the highest stream ID processed.

src/Driver/Http2Driver.php
trowski: Maybe this isn't necessary anyway, since many small window updates and small frames should be caught by the limits on total bytes vs. payload bytes. Let me give that some thought.

src/Driver/Http2Driver.php
trowski: Being in the parser, `yield` was causing `Client::close()` to not be called unless more data was received from the client (though a destructor probably called it anyway). Fixed in master as it isn't really relevant to this PR.

trowski:
Rebased on master.

src/Driver/Http2Driver.php
kelunik: Not sure whether self::ENHANCE_YOUR_CALM is the correct error code here?

PR #289, opened 2019-10-11
Make connection timeout driver dependent

trowski:


src/Driver/Http1Driver.php
```diff
@@ -108,6 +113,15 @@ public function getPendingRequestCount(): int
         return 0;
     }
 
+    public function getExpirationTime(): int
+    {
+        if ($this->http2) {
+            return $this->http2->getExpirationTime();
+        }
+
+        return $this->timeReference->getCurrentTime() + $this->timeout;
```
kelunik: Shouldn't this be _lastClientRead_ or something?

trowski: It's not that simple. It's also based on last write.

src/Options.php
```diff
@@ -7,7 +7,9 @@ final class Options
     private $debug = false;
     private $connectionLimit = 10000;
     private $connectionsPerIpLimit = 30; // IPv4: /32, IPv6: /56 (per RFC 6177)
-    private $connectionTimeout = 15; // seconds
+    private $keepAliveTimeout = 15; // seconds
```
kelunik: This should be `http1Timeout` for consistency.

src/Driver/Http1Driver.php
kelunik: But it shouldn't be `$now + $timeout` here, no?

trowski: Sorry, for HTTP/1.x it is simple – add whatever the keep-alive timeout is to the last time the was activity on the connection.

trowski: Clarification: `getExpirationTime` is called when there is a read/write on the socket. For HTTP/1.x we can simply add the keep-alive timeout. For HTTP/2, because we can receive control frames, the expiration time is not simple – it's only updated within the driver when payload data is received/written.

kelunik: That's pretty much my concern: The interface says:

```php
/**
 * @return int Timestamp when the connection should be closed. This time should be updated when data is sent
 *             or received on the connection.
 */
public function getExpirationTime(): int; 
```

The `Http2Driver` does exactly that, while the `Http1Driver` returns a value based on the current time and the timeout, making the return value call time dependent and moving the responsibility to call it at the right time into `Client`.

PR #292, opened 2019-12-10
Fix bug related to HTTP content chunking

bzikarsky:
The current implementation of chunk parsing can break in specific combinations of HTTP chunk-size and options::$chunkSize.

I added a test-data provider which runs the chunk test with chunk-sizes from 1 to 10 against the example HTTP request with chunks in size 4 and 5. This breaks without the fix to `Http1Driver`.

I also had to edit `composer.json` to make sure PHPUnit does not upgrade to v8 as this is incompatible with the listener configuration in `phpunit.xml.dist`. I considered it out of the scope of this MR to fix the deprecated listener config.

### DIFF
--- a/test/Driver/Http1DriverTest.php
+++ b/test/Driver/Http1DriverTest.php
@@ -20,6 +20,7 @@ use Amp\Http\Status;
 use Amp\PHPUnit\TestCase;
 use Amp\Promise;
 use Amp\Success;
+use Generator;
 use League\Uri;
 
 class Http1DriverTest extends TestCase
@@ -218,7 +219,21 @@ class Http1DriverTest extends TestCase
         $this->assertSame($originalBody, $body);
     }
 
-    public function testChunkedBodyParseEmit()
+    /**
+     * provide multiple chunk-sizes to test with.
+     * @return Generator
+     */
+    public function chunkSizeProvider()
+    {
+        for ($i = 1; $i < 11; $i++) {
+            yield [$i];
+        }
+    }
+
+    /**
+     * @dataProvider chunkSizeProvider
+     */
+    public function testChunkedBodyParseEmit(int $chunkSize)
     {
         $msg =
             "POST https://test.local:1337/post-endpoint HTTP/1.0\r\n" .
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -47,7 +47,7 @@
         "amphp/php-cs-fixer-config": "dev-master",
         "league/uri-components": "^1.7",
         "monolog/monolog": "^1.23",
-        "phpunit/phpunit": "^6",
+        "phpunit/phpunit": "^6,<8",
```
kelunik: `^6` does already forbid installing `8`, it's equivalent to `>=6 && < 7`.

test/Driver/Http1DriverTest.php
```diff
@@ -251,8 +266,8 @@ public function testChunkedBodyParseEmit()
             $this->createCallback(0)
         );
 
-        for ($i = 0, $c = \strlen($msg); $i < $c; $i++) {
-            $promise = $parser->send($msg[$i]);
+        foreach (str_split($msg, $chunkSize) as $chunk) {
+            $promise = $parser->send($chunk);
         }
 
         while ($promise instanceof Promise) {
```
kelunik: This while should probably be inside the `foreach`, @trowski?

composer.json
bzikarsky: I thought so as well, but composer somehow decided to provide me with PHPunit 8 in the v1.x branch.

bzikarsky: Strange, maybe a composer.lock from master spilled over. After cleaning everything, PHPunit was correctly pulled. I'll update the PR accordingly.

test/Driver/Http1DriverTest.php
bzikarsky: Do you want me to change this @kelunik without @trowski's feedback? It looks like it'd make sense and the tests (still) pass with the `while`-loop moved inside.

kelunik: Yes, please.

bzikarsky: Done

trowski: Yep, definitely belonged in the `foreach`.

bzikarsky:
I just updated the MR to fix the styling issues.

bzikarsky:
Thanks for merging - Can we also get a bugfix release for this change? :-)

trowski:
Sorry, meant to do this after tagging 2.0-RC4. Tagged now!

bzikarsky:
Awesome, thanks!

kelunik:
@trowski Did you merge this into 2.x?

trowski:
@kelunik Yes I did.

PR #300, opened 2020-02-06
Make raw HTTP/1.x headers available in Request

trowski:
Responses are still written with lower-case headers, but this PR modifies the HTTP/1.x parser to make the original casing of Request headers available.

### DIFF
--- a/src/Driver/Http1Driver.php
+++ b/src/Driver/Http1Driver.php
@@ -344,7 +344,11 @@ final class Http1Driver implements HttpDriver
                         $rawHeaders = \preg_replace(Rfc7230::HEADER_FOLD_REGEX, ' ', $rawHeaders);
                     }
 
-                    $headers = Rfc7230::parseHeaders($rawHeaders);
+                    $headers = Rfc7230::parseRawHeaders($rawHeaders);
+                    $headerMap = [];
+                    foreach ($headers as [$key, $value]) {
+                        $headerMap[\strtolower($key)][] = $value;
+                    }
                 } catch (InvalidHeaderException $e) {
                     throw new ClientException(
                         $this->client,
@@ -361,7 +365,7 @@ final class Http1Driver implements HttpDriver
                     );
                 }
 
-                $contentLength = $headers["content-length"][0] ?? null;
+                $contentLength = $headerMap["content-length"][0] ?? null;
                 if ($contentLength !== null) {
                     if (!\preg_match("/^(?:0|[1-9][0-9]*)$/", $contentLength)) {
                         throw new ClientException(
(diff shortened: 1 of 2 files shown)


src/Driver/Http1Driver.php
```diff
@@ -475,25 +479,25 @@ private function parser(): \Generator
                     );
                 }
 
-                if (isset($headers["expect"][0]) && \strtolower($headers["expect"][0]) === "100-continue") {
+                if (isset($headerMap["expect"][0]) && \strtolower($headerMap["expect"][0]) === "100-continue") {
                     yield $this->write(
-                        new Request($this->client, $method, $uri, $headers, null, $protocol),
+                        new Request($this->client, $method, $uri, $headerMap, null, $protocol),
```
kelunik: Shouldn't this also use the `$headers` variables and add the headers manually?

src/Driver/Http1Driver.php
```diff
@@ -475,25 +479,25 @@ private function parser(): \Generator
...
-                    && \strtolower($headers["upgrade"][0]) === "h2c"
-                    && false !== $h2cSettings = \base64_decode(\strtr($headers["http2-settings"][0], "-_", "+/"), true)
+                    && false !== \stripos($headerMap["connection"][0], "upgrade")
+                    && \strtolower($headerMap["upgrade"][0]) === "h2c"
+                    && false !== $h2cSettings = \base64_decode(\strtr($headerMap["http2-settings"][0], "-_", "+/"), true)
                 ) {
                     // Request instance will be overwritten below. This is for sending the switching protocols response.
                     yield $this->write(
-                        new Request($this->client, $method, $uri, $headers, null, $protocol),
+                        new Request($this->client, $method, $uri, $headerMap, null, $protocol),
```
kelunik: Shouldn't this also use the `$headers` variables and add the headers manually?

src/Driver/Http1Driver.php
trowski: This `Request` instance is only used internally, it is never given to a `RequestHandler`, so it doesn't matter since the original header case is never used.

src/Driver/Http1Driver.php
trowski: Same as above.

src/Driver/Http1Driver.php
```diff
@@ -522,15 +528,14 @@ private function parser(): \Generator
...
-                        $protocol
-                    ), $buffer);
 
+                    $request = new Request($this->client, $method, $uri, [], null, $protocol);
+                    foreach ($headers as [$key, $value]) {
+                        $request->addHeader($key, $value);
+                    }
+
+                    yield $this->pendingResponse = ($this->onMessage)($request, $buffer);
+                    $request = null; // DO NOT leave a reference to the Request object in the parser!
```
kelunik: Should we `null` the request here before yielding?

trowski:
Manually merged.

PR #305, opened 2020-04-02
Add CORS links

enumag:
As requested [here](https://github.com/amphp/http-server/issues/303#issuecomment-594734312).


kelunik:
Thanks!

PR #306, opened 2020-04-05
Improve exception logging

enumag:
I had a problem that my monolog handlers were not logging exceptions correctly. It is because they expect the exception to be in the context.

### DIFF
--- a/src/Driver/RemoteClient.php
+++ b/src/Driver/RemoteClient.php
@@ -610,7 +610,7 @@ final class RemoteClient implements Client
             return;
         } catch (\Throwable $exception) {
             $errorType = \get_class($exception);
-            $this->logger->critical(
+            $this->logger->error(
                 "Unexpected {$errorType} thrown from RequestHandler::handleRequest(), falling back to error handler.",
                 ['exception' => $exception]
             );
@@ -668,7 +668,7 @@ final class RemoteClient implements Client
         } catch (\Throwable $exception) {
             // If the error handler throws, fallback to returning the default error page.
             $errorType = \get_class($exception);
-            $this->logger->critical(
+            $this->logger->error(
                 "Unexpected {$errorType} thrown from ErrorHandler::handleError(), falling back to default error handler.",
                 ['exception' => $exception]
             );
@@ -705,7 +705,7 @@ final class RemoteClient implements Client
             }
 
             $errorType = \get_class($exception);
-            $this->logger->critical(
+            $this->logger->error(
                 "Unexpected {$errorType} thrown during socket upgrade, closing connection.",
                 ['exception' => $exception]
             );
--- a/src/Driver/RemoteClient.php
+++ b/src/Driver/RemoteClient.php
@@ -605,7 +605,7 @@ final class RemoteClient implements Client
             $this->close();
             return;
         } catch (\Throwable $exception) {
-            $this->logger->error($exception->getMessage(), ['exception' => $exception]);
+            $this->logger->error('An exception was thrown while processing a request.', ['exception' => $exception]);
             $response = yield from $this->makeExceptionResponse($request);
         } finally {
             $this->pendingHandlers--;
--- a/src/Driver/RemoteClient.php
+++ b/src/Driver/RemoteClient.php
@@ -691,7 +691,7 @@ final class RemoteClient implements Client
                 return;
             }
 
-            $this->logger->error($exception->getMessage(), ['exception' => $exception]);
+            $this->logger->error('Unexpected exception during socket upgrade, closing connection.', ['exception' => $exception]);
             $this->close();
         });
     }
--- a/src/Driver/RemoteClient.php
+++ b/src/Driver/RemoteClient.php
@@ -658,7 +658,7 @@ final class RemoteClient implements Client
             return yield $this->errorHandler->handleError($status, null, $request);
         } catch (\Throwable $exception) {
             // If the error handler throws, fallback to returning the default HTML error page.
-            $this->logger->error($exception->getMessage(), ['exception' => $exception]);
+            $this->logger->error('Error handler threw an exception, falling back to the default HTML error page.', ['exception' => $exception]);
 
             // The default error handler will never throw, otherwise there's a bug
             return yield self::$defaultErrorHandler->handleError($status, null, $request);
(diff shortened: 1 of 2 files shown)


enumag:
By the way, psr/log actually states that the first argument should be a string, see https://github.com/php-fig/log/blob/master/Psr/Log/LoggerInterface.php. By doing this you're violating the contract. I strongly recommend adding PHPStan to your CI to discover such issues.

Also please tell me if you're using Logger in any other amphp package so that I can fix it everywhere.

src/Driver/RemoteClient.php
```diff
@@ -396,7 +396,7 @@ private function parse(?string $data = null): void
             }
         } catch (\Throwable $exception) {
             // Parser *should not* throw an exception, but in case it does...
-            $this->logger->critical($exception);
+            $this->logger->critical($exception->getMessage(), ['exception' => $exception]);
```
kelunik, line 399: This should probably cast the exception to a string?

enumag: I don't think so. I checked Symfony to see how they're doing it and it turns out that they're not even using the exception message. They always use some string literal for instance:

```php
$this->logger->error('An exception was thrown when handling an AccessDeniedException.', array('exception' => $e));
```

So it would actually be better to not even use the message.

src/Middleware/ExceptionMiddleware.php
```diff
@@ -41,7 +41,7 @@ public function handleRequest(Request $request, RequestHandler $requestHandler):
             } catch (\Throwable $exception) {
                 $status = Status::INTERNAL_SERVER_ERROR;
 
-                $this->logger->error($exception);
+                $this->logger->error($exception->getMessage(), ['exception' => $exception]);
```
enumag, line 44:
```suggestion
                $this->logger->error('An exception was thrown while processing a request, causing an error 500 response.', ['exception' => $exception]);
```

src/Driver/RemoteClient.php
enumag, line 399:
```suggestion
            $this->logger->critical('Request parser threw an exception (this should never happen).', ['exception' => $exception]);
```

src/Driver/RemoteClient.php
```diff
@@ -605,7 +605,7 @@ private function respond(Request $request, string $buffer): \Generator
             $this->close();
             return;
         } catch (\Throwable $exception) {
-            $this->logger->error($exception);
+            $this->logger->error($exception->getMessage(), ['exception' => $exception]);
```
enumag, line 608:
```suggestion
            $this->logger->error('An exception was thrown while processing a request.', ['exception' => $exception]);
```

enumag: I'm not really satisfied with this and the case in ExceptionMiddleware. I think the messages should be different.

src/Driver/RemoteClient.php
```diff
@@ -658,7 +658,7 @@ private function makeExceptionResponse(Request $request): \Generator
             return yield $this->errorHandler->handleError($status, null, $request);
         } catch (\Throwable $exception) {
             // If the error handler throws, fallback to returning the default HTML error page.
-            $this->logger->error($exception);
+            $this->logger->error($exception->getMessage(), ['exception' => $exception]);
```
enumag, line 661:
```suggestion
            $this->logger->error('Error handler threw an exception, falling back to the default HTML error page.', ['exception' => $exception]);
```

src/Driver/RemoteClient.php
```diff
@@ -691,7 +691,7 @@ private function export(callable $upgrade, Request $request, Response $response,
                 return;
             }
 
-            $this->logger->error($exception);
+            $this->logger->error($exception->getMessage(), ['exception' => $exception]);
```
enumag, line 694: I have no idea what this is about... can you write some message for this please?

enumag:
ping @kelunik I'd like to get this sorted out

enumag:
Bump

enumag:
ping @trowski @kelunik

src/Driver/RemoteClient.php
kelunik: In that case this shouldn't use the message, but rather say something like the comment above says: `'Unexpected ' . \get_class($e) . ' while parsing data from client @ ' . (string) $this->clientAddress`

src/Driver/RemoteClient.php
kelunik: Whether it's an HTML error page depends on the error handler.

src/Driver/RemoteClient.php
kelunik: `Unexpected exception during socket upgrade, closing connection`

src/Driver/RemoteClient.php
enumag: See https://github.com/amphp/http-server/pull/306#discussion_r403727293. Gimme message suggestions for the individual logger calls below if you don't like mine.

src/Driver/RemoteClient.php
enumag: ? I mean this is a case where the error handler threw so I don't think it depends on it anymore.

src/Driver/RemoteClient.php
enumag:
```suggestion
            $this->logger->error('Unexpected exception during socket upgrade, closing connection.', ['exception' => $exception]);
```

kelunik:
@enumag I've applied all suggestions and changed the exception messages to what I think is good. Could you please have a look again, before I merge it?

Sorry for the delay, I always read the pings on my phone and then always forgot about it.

src/Driver/RemoteClient.php
```diff
@@ -396,7 +396,11 @@ private function parse(?string $data = null): void
             }
         } catch (\Throwable $exception) {
             // Parser *should not* throw an exception, but in case it does...
-            $this->logger->critical($exception);
+            $errorType = \get_class($exception);
+            $this->logger->critical("Unexpected {$errorType} while parsing request, closing connection.", [
+                'exception' => $exception
+            ]);
```
enumag, line 402:
```suggestion
            $this->logger->critical(
                "Unexpected {$errorType} while parsing request, closing connection.",
                ['exception' => $exception]
            );
```

enumag:
@kelunik Thanks for finding the time. Looks good to me. Some tests are failing but I highly doubt it's related to this PR.

PR #310, opened 2020-05-07
Pass Request to Logger

### LINKED ISSUE #309: Sending additional data to logger
When sending errors to an error tracking software such as Sentry, it is nice to be able to attach some additional data when `\Psr\Log\LoggerInterface` is called. 

In particular:
- Some information about the request itself.
- Information about the authenticated user.

Since these data are specific to each request it needs to be somehow handled by the HttpServer and the data need to be added to the `$context` array when logger is called. I'd like to ask for recommendations how to implement this.

First thing I'm wondering about is the `ExceptionMiddleware` - it is clearly something that takes care of logging the exceptions and sending appropriate html to the client. However it isn't used by default, nor mentioned anywhere in the documentation. So what is different in this middleware from the default behavior? Where in the [stack](https://github.com/amphp/http-server/blob/10b29b19e3900a5ebe251a139d6c04834cc25e9c/src/Middleware/functions.php#L17) should I use it?

Second thing is integrating user authentication. I assume I should do it as a middleware and add the user data to the request using `Request::setAttribute()` so that they can be available in both the final RequestHandler and in other middlewares (for logging or something). Does that sound correct?

Third question is about modifying the context when logger is called. Even the ExceptionMiddleware only passes the exception but nothing else. Should I implement my own version of it or should we try to improve the existing one with a callback `Closure(Request, Exception): array` or something?

enumag:
Closes #309

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -32,7 +32,7 @@ You can then access the example server at [`http://localhost:1337/`](http://loca
 The `Server` uses a `NullLogger` by default.
 If you pass a `Psr\Log\LoggerInterface` instance to its constructor, you'll get helpful log messages.
 
-Optionally you can use `Options::withRequestLogging()` to enable passing the `Request` object to the logger in the `$context` array. This is disabled by default.
+Optionally you can use `Options::withRequestLogContext()` to enable passing the `Request` object to the logger in the `$context` array. This is disabled by default.
 
 {:.note}
 > Internally generated log messages of the `DEBUG` level are only generated if `zend.assertions` is set to `1`.
--- a/docs/index.md
+++ b/docs/index.md
@@ -32,6 +32,8 @@ You can then access the example server at [`http://localhost:1337/`](http://loca
 The `Server` uses a `NullLogger` by default.
 If you pass a `Psr\Log\LoggerInterface` instance to its constructor, you'll get helpful log messages.
 
+Optionally you can use `Options::withRequestLogging()` to enable passing the `Request` object to the logger in the `$context` array. This is disabled by default.
+
 {:.note}
 > Internally generated log messages of the `DEBUG` level are only generated if `zend.assertions` is set to `1`.
 > If it isn't set to `1` in your config, load the server with `php -d zend.assertions=1 examples/hello-world.php`.
--- a/src/Options.php
+++ b/src/Options.php
@@ -23,6 +23,7 @@ final class Options
     private $compression = true;
     private $allowHttp2Upgrade = false;
     private $pushEnabled = true;
+    private $requestLogging = false;
 
     /**
      * @return bool `true` if server is in debug mode, `false` if in production mode.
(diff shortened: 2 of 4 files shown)


kelunik:
I guess we have to make it opt-in, otherwise sensitive information might be logged by accident.

bwoebi:
Technically the wrong exceptions may also end up containing sensitive information. But nevertheless, should be probably enabled by default if `Options->debug` is true.

enumag:
Logging tools for standard PHP often take the request info from superglobal variables and log it by default. And as bwoebi said, exceptions can contain sensitive information too. In my opinion this shouldn't be a concern of this library. There are always ways to remove these things from the log by configuring the logging tool as needed.

kelunik:
@enumag Enabling it by default is maybe fine in a major version, but we like to keep things secure by default. Almost any request will contain either `authorization` headers or `cookie` headers. Exceptions _might_ contain sensitive information, but for requests this is almost _guaranteed_.

enumag:
I decided I'd test this first before making it optional. And I discovered another issue which I'd like to discuss.

As noted in #309, I refactored my authentication to a middleware and the user is inserted into Request attributes. As such, I expected to get the user information in the logging layer as well because I would like to log which user sent the request.

At the moment this doesn't work however since the Request instance my middleware is working with is different from the one that is logged because of this [clone](https://github.com/amphp/http-server/blob/10b29b19e3900a5ebe251a139d6c04834cc25e9c/src/Driver/RemoteClient.php#L598). What is the reason for cloning the request? If it is necessary for some reason, would it be possible for the logger to get the cloned and middleware-altered instance?

enumag:
As for making it optional - how should I implement that? Is it alright to introduce a new option in the [Options](https://github.com/amphp/http-server/blob/master/src/Options.php) class? I don't want it to be tied to the debug option since I'd like to log Request in production.

enumag:
ping @kelunik

kelunik:
I'd add it to the `Options` class, yes.

trowski:
> At the moment this doesn't work however since the Request instance my middleware is working with is different from the one that is logged because of this clone.

The clone is so `HttpDriver::write()` receives the unmodified request object. You can add `$clientRequest = clone $request` at the top of `respond()` and pass `$clientRequest` to `write()`. Using the modified request everywhere else should be fine, including the upgrade handler, which really should have been getting the user-modified request.

enumag:
Thanks guys, I'll implement it.

enumag:
@kelunik @trowski All done.

enumag:
Added a note to documentation as well. ;-)

src/Options.php
```diff
@@ -513,4 +514,31 @@ public function withoutCompression(): self
...
+
+    public function isRequestLoggingEnabled(): bool
+    {
+        return $this->requestLogging;
+    }
+
+    /**
+     * Enables passing the causing Request object to Logger.
+     */
+    public function withRequestLogging(): self
```
kelunik, line 526: This sounds more like an access log option. Maybe `withRequestLogContext`?

docs/index.md
```diff
@@ -32,6 +32,8 @@ You can then access the example server at [`http://localhost:1337/`](http://loca
 The `Server` uses a `NullLogger` by default.
 If you pass a `Psr\Log\LoggerInterface` instance to its constructor, you'll get helpful log messages.
 
+Optionally you can use `Options::withRequestLogging()` to enable passing the `Request` object to the logger in the `$context` array. This is disabled by default.
```
kelunik, line 35:
```suggestion
Optionally you can use `Options::withRequestLogContext()` to enable passing the `Request` object to the logger in the `$context` array. This is disabled by default.
```

kelunik:
Thanks!

enumag:
Thanks! Can I ask for a release soon-ish?

PR #314, opened 2020-06-18
Discard request bodies on HTTP/2 if not consumed

kelunik:


kelunik:
Nevermind, the body is already discarded by `RequestBody`, as it extends `Payload`. However, I've separately applied a fix for too large bodies via https://github.com/amphp/http-server/commit/30063e245559095ff99a320b88a8a88ff719cc6e.

PR #315, opened 2020-09-03
Implemented listenAndServe, handleFunc and html functions

mnavarrocarter:
Added a couple of functions inspired in golang's http package.

`listenAndServe` makes really quick and straightforward to boot up a server passing a handler and a socket address.

`handleFunc` is a convenience function to transform a callable into a handler. It is less verbose that constructing the adapter.

`html` is just a wrapper for creating an html response.

Overall, these functions make the quickstart process easier, more elegant, more readable, and a bit more "golangish". 😛 

I modified the readme showcasing the simplified approach:

```php
<?php

use Amp\Http\Server\Request;
use Amp\Promise;
use Amp\Success;
use function Amp\Http\Server\handleFunc;
use function Amp\Http\Server\html;
use function Amp\Http\Server\listenAndServe;

// Run this script, then visit http://localhost:8000/ in your browser.

function helloWorld(Request $request): Promise {
    return new Success(html('Hello World!'));
}

Amp\Loop::run(static function () {
    yield from listenAndServe('0.0.0.0:8000', handleFunc('helloWorld'));
});

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,36 +40,21 @@ composer require amphp/http-server
 ```php
 <?php
 
-use Amp\Http\Server\RequestHandler\CallableRequestHandler;
-use Amp\Http\Server\HttpServer;
 use Amp\Http\Server\Request;
 use Amp\Http\Server\Response;
-use Amp\Http\Status;
-use Amp\Socket\Server;
-use Psr\Log\NullLogger;
+use function Amp\Http\Server\handleFunc;
+use function Amp\Http\Server\html;
+use function Amp\Http\Server\listenAndServe;
 
-// Run this script, then visit http://localhost:1337/ in your browser.
+// Run this script, then visit http://localhost:8000/ in your browser.
 
-Amp\Loop::run(function () {
-    $sockets = [
-        Server::listen("0.0.0.0:1337"),
-        Server::listen("[::]:1337"),
-    ];
+function helloWorld(Request $request): Response
+{
+    return html('Hello World!');
+}
 
-    $server = new HttpServer($sockets, new CallableRequestHandler(function (Request $request) {
-        return new Response(Status::OK, [
-            "content-type" => "text/plain; charset=utf-8"
-        ], "Hello, World!");
-    }), new NullLogger);
-
-    yield $server->start();
-
-    // Stop the server gracefully when SIGINT is received.
-    // This is technically optional, but it is best to call Server::stop().
-    Amp\Loop::onSignal(SIGINT, function (string $watcherId) use ($server) {
-        Amp\Loop::cancel($watcherId);
-        yield $server->stop();
-    });
+Amp\Loop::run(static function () {
+    yield listenAndServe('0.0.0.0:8000', handleFunc('helloWorld'));
 });
 ```
 
--- a/examples/hello-world-simple.php
+++ b/examples/hello-world-simple.php
@@ -1,0 +1,18 @@
+#!/usr/bin/env php
+<?php
+require \dirname(__DIR__) . "/vendor/autoload.php";
+
+use Amp\Http\Server\Request;
+use Amp\Http\Server\Response;
+use function Amp\Http\Server\handleFunc;
+use function Amp\Http\Server\html;
+use function Amp\Http\Server\listenAndServe;
+
+function helloWorld(Request $request): Response
+{
+    return html('Hello World!');
+}
+
+Amp\Loop::run(static function () {
+    yield listenAndServe('0.0.0.0:8000', handleFunc('helloWorld'));
+});
--- a/src/functions.php
+++ b/src/functions.php
@@ -2,9 +2,65 @@
 
 namespace Amp\Http\Server;
 
+use Amp\ByteStream\InputStream;
 use Amp\Http\Status;
+use Amp\Promise;
+use Amp\Socket\Server as Socket;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use function Amp\call;
 
 function redirectTo(string $uri, int $statusCode = Status::FOUND): Response
 {
     return new Response($statusCode, ['location' => $uri]);
+}
+
+/**
+ * Returns a response with the html content type.
+ *
+ * @param InputStream|string|null $content
+ */
+function html($content, int $status = Status::OK): Response
+{
+    return new Response($status, [
+        'Content-Type' => 'text/html; charset=utf-8'
+    ], $content);
+}
+
+/**
+ * Creates an HTTP server.
+ *
+ * This function is inspired in Golang's http module one. It provides sensible defaults for a really quick start
+ * and it is designed to cover 90% of use cases.
+ *
+ * If you need fine-tuning, like listening in two sockets or attach more listeners to the server
+ * you should bootstrap your server manually.
+ *
+ * @return Promise<HttpServer>
+ */
+function listenAndServe(string $address, RequestHandler $handler, ?Options $options, ?LoggerInterface $logger): Promise
+{
+    return call(static function () use ($address, $handler, $options, $logger) {
+        $sockets = [
+            Socket::listen($address)
+        ];
+
+        $logger = $logger ?? new NullLogger();
+
+        $server = new HttpServer($sockets, $handler, $logger, $options);
+
+        yield $server->start();
+
+        return $server;
+    });
+}
+
+/**
+ * Wraps a php callable in an amp http server handler.
+ *
+ * @param callable(Request): Promise<Response> $callable
+ */
+function handleFunc(callable $callable): RequestHandler
+{
+    return new RequestHandler\CallableRequestHandler($callable);
 }


src/functions.php
```diff
@@ -2,9 +2,77 @@
...
+
+    $server = new HttpServer($sockets, $handler, $logger, $options);
+
+    yield $server->start();
+
+    // Call to server stop on SIGINT for a graceful stop
+    Loop::onSignal(SIGINT, static function ($watcherId) use ($server) {
+        Loop::cancel($watcherId);
+        yield $server->stop();
+    });
```
kelunik, line 66: I don't like the signal handler being registered here. It breaks on windows and you probably want sigint and sigterm. The signal is caught here and the server stopped, but the program doesn't exit.

kelunik (commented):
I like the general idea, but there are some things that need to be changed, which I have commented separately.

composer.json
```diff
@@ -30,6 +30,7 @@
     ],
     "require": {
         "php": ">=7.2",
+        "ext-pcntl": "*",
```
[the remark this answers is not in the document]
mnavarrocarter: [The SIGINT constant belongs to the pcntl extension](https://www.php.net/manual/en/pcntl.constants.php). In some environments that try to keep size at the minimum (i.e. alpine docker images) it may not be present. If we require it with composer we ensure the package cannot be installed if the environment does not support handling of signals. 

Of course I'm assuming that's the behavior you want. I can totally remove it.

src/functions.php
mnavarrocarter: Mmm, so I need a way to pass the created `HttpServer` to the client code calling the `listenAndServe` function so they can register the signal handlers they need, but I'm unsure how to do that. Any guidance?

src/functions.php
```diff
@@ -2,9 +2,77 @@
...
+ *
+ * @param string $address
+ * @param RequestHandler $handler
+ * @param Options|null $options
+ * @param LoggerInterface|null $logger
+ * @return Generator
+ * @throws Loop\UnsupportedFeatureException
+ * @throws SocketException
+ */
+function listenAndServe(string $address, RequestHandler $handler, Options $options = null, LoggerInterface $logger = null): Generator
```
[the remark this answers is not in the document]
mnavarrocarter: Great, I'll fix it.

examples/hello-world-simple.php
```diff
@@ -0,0 +1,19 @@
...
+use Amp\Http\Server\Request;
+use Amp\Promise;
+use Amp\Success;
+use function Amp\Http\Server\handleFunc;
+use function Amp\Http\Server\html;
+use function Amp\Http\Server\listenAndServe;
+
+function helloWorld(Request $request): Promise
+{
+    return new Success(html('Hello World!'));
```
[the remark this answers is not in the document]
mnavarrocarter: Great, I'll fix it.

src/functions.php
enumag: @mnavarrocarter `return $server` ?

enumag:
```php
/**
 * @return Promise<HttpServer>
 */
function listenAndServe(string $address, RequestHandler $handler, Options $options = null, LoggerInterface $logger = null): Promise
{
    return call(function () use (...) {
        // your current code

        return $server;
    });
}

File: composer.json (hunk 2 above)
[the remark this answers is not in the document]
Comment (mnavarrocarter) [reply composer.json:33]: Well, if I remove the signal handling, I guess I don't need to include it.

File: src/functions.php [hunk 5]
```diff
@@ -2,9 +2,77 @@
...
+    Loop::onSignal(SIGINT, static function ($watcherId) use ($server) {
+        Loop::cancel($watcherId);
+        yield $server->stop();
+    });
+}
+
+/**
+ * Wraps a php callable in an amp http server handler.
+ *
+ * @param callable $callable
```
enumag, line 72: document the callable with parameter types and return type like `@param callable(ParamType): ReturnType $callable`

src/functions.php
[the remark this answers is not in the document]
enumag: You could allow `$address` to be an array and listen to multiple sockets that way. Thoughts @kelunik ?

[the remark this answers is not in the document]
enumag: Also for options and logger I'd add question marks to the typehints.

Personally I'd also drop the `@param` annotations since they don't add anything and only duplicate the typehints anyway. But then again current Amphp code usually has those duplications so maybe @trowski and @kelunik want them?

src/functions.php
mnavarrocarter: I didn't know you could do that. Is that psalm?

enumag: it's common syntax these days... supported by both phpstan and psalm, quite possibly by other tools as well

src/functions.php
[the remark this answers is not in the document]
kelunik: We usually only have them if there's also a description or more specific types.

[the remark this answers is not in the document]
mnavarrocarter: > You could allow `$address` to be an array and listen to multiple sockets that way. Thoughts @kelunik ?

Most of the servers I've implemented just listen in one socket. But I'm not against the idea. I think the function will work for the vast majority of use cases.

mnavarrocarter:
I think the last update I pushed should solve all the issues.

Thanks a lot for the feedback! 😄

src/functions.php
mnavarrocarter: Thanks, still wrapping my head around coroutines.

src/functions.php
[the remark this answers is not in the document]
enumag: > Most of the servers I've implemented just listen in one socket.

These days I'd argue to listen on both IPv4 and IPv6.

README.md
```diff
@@ -40,36 +40,21 @@ composer require amphp/http-server
...
+
+// Run this script, then visit http://localhost:8000/ in your browser.
+
+function helloWorld(Request $request): Response
+{
+    return html('Hello World!');
+}
+
+Amp\Loop::run(static function () {
+    yield listenAndServe('0.0.0.0:8000', handleFunc('helloWorld'));
```
kelunik, line 57: Misses the shutdown now.

src/functions.php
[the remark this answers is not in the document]
enumag: @kelunik Usually eh? Idk about that... [this](https://github.com/enumag/http-server/commit/fdddbfbf17a0c02bb3bacbef0189dbeb1cb5ceb6) is just one file... Can I go and wipe them out?

mnavarrocarter:
Last update. I removed default `null` arguments and fallback to the `?` optional type.

src/functions.php
```diff
@@ -2,9 +2,72 @@
...
 function redirectTo(string $uri, int $statusCode = Status::FOUND): Response
 {
     return new Response($statusCode, ['location' => $uri]);
 }
+
+/**
+ * Returns a response with the html content type.
+ *
+ * @param InputStream|string|null $content
+ * @param int $status
```
enumag, line 22:
```suggestion
```

src/functions.php
```diff
@@ -2,9 +2,72 @@
...
+ * This function is inspired in Golang's http module one. It provides sensible defaults for a really quick start
+ * and it is designed to cover 90% of use cases.
+ *
+ * If you need fine-tuning, like listening in two sockets or attach more listeners to the server
+ * you should bootstrap your server manually.
+ *
+ * @param string $address
+ * @param RequestHandler $handler
+ * @param Options|null $options
+ * @param LoggerInterface|null $logger
```
enumag, line 44:
```suggestion
```

src/functions.php
```diff
@@ -2,9 +2,72 @@
...
+
+        return $server;
+    });
+}
+
+/**
+ * Wraps a php callable in an amp http server handler.
+ *
+ * @param callable(Request): Promise<Response> $callable
+ * @return RequestHandler
```
enumag, line 68:
```suggestion
```

enumag (changes requested):

enumag:
> Last update. I removed default null arguments and fallback to the ? optional type.

The default null value is probably a good thing in this case. It's just generally preferred to have the question mark in place even together with `= null`.

src/functions.php
```diff
@@ -2,9 +2,72 @@
...
+    return new Response($status, [
+        'Content-Type' => 'text/html; charset=utf-8'
+    ], $content);
+}
+
+/**
+ * Creates an HTTP server.
+ *
+ * This function is inspired in Golang's http module one. It provides sensible defaults for a really quick start
+ * and it is designed to cover 90% of use cases.
```
kelunik, line 34: It won't cover 90% I guess. It only listens with amphp/socket, while generally amphp/cluster will be used to allow for multi processing with multiple CPU cores.

src/functions.php
```diff
@@ -2,9 +2,72 @@
...
+ * @return Promise<HttpServer>
+ */
+function listenAndServe(string $address, RequestHandler $handler, ?Options $options, ?LoggerInterface $logger): Promise
+{
+    return call(static function () use ($address, $handler, $options, $logger) {
+        $sockets = [
+            Socket::listen($address)
+        ];
+
+        $logger = $logger ?? new NullLogger();
```
kelunik, line 48: I think I'd add logger before options and not make it optional, because it's really helpful to have it enabled. Especially if you're starting and don't know what might be wrong.

mnavarrocarter:
Okay, `= null` is back again. I removed the redundant annotation declarations too.

src/functions.php
```diff
@@ -2,9 +2,65 @@
...
+        return $server;
+    });
+}
+
+/**
+ * Wraps a php callable in an amp http server handler.
+ *
+ * @param callable(Request): Promise<Response> $callable
+ */
+function handleFunc(callable $callable): RequestHandler
```
kelunik, line 63: ListenAndServe could also accept callable | RequestHandler.

src/functions.php
mnavarrocarter: Yeah, well, we usually have a single core in our Kubernetes pods and we scale spawning more processes. But I see what you mean.

Willing to change the description though. 😛

src/functions.php
mnavarrocarter: Well, I thought that at the beginning, but people end up passing the `NullLogger` anyways for a quick start. The optional method does not prevent you to use a logger for debugging if you need to, but it does make the quickstart simple. 😄 

And yeah, I think logger before options is actually better. Found myself passing the logger more than the options.

README.md
mnavarrocarter: Oops. Sorry!

mnavarrocarter: I just need to do `return yield listenAndServe()` right?

mnavarrocarter: Not quite sure if I get what you mean.

PR #316, opened 2020-09-03
Remove useless phpdoc tags

enumag:
Should I continue?

### DIFF
--- a/src/HttpServer.php
+++ b/src/HttpServer.php
@@ -82,8 +82,6 @@ final class HttpServer
 
     /**
      * @param SocketServer[] $servers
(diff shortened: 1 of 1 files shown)


mnavarrocarter:
Maybe would be better to add the [`no_superfluous_phpdoc_tags`](https://mlocati.github.io/php-cs-fixer-configurator/#version:2.16|fixer:no_superfluous_phpdoc_tags) to the Amp Code Style to make it official?

enumag:
Great suggestion! It should even be able to remove it automatically.

kelunik:
Thanks!

PR #325, opened 2021-12-19
Fixes on v3

iggyvolz:
Made a few broad fixes on the v3 branch that get the examples running and most of the tests passing:
* `Socket\Server::listen` => `Socket\listen` as per https://github.com/amphp/socket/commit/88b0e8dee4d9eece2fd53deb6bcb088fa636a416
* `PipelineStream(new AsyncGenerator` => `IterableStream` as per https://github.com/amphp/pipeline/releases/tag/v1.0.0-beta.2 (not 100% sure if I did this right but it seems to work)
* `EmmitterStream::asPipeline()` `EmmitterStream::pipe()`
* `InMemoryStream` => `ReadableBuffer` as per https://github.com/amphp/byte-stream/releases/tag/v2.0.0-beta.1

### DIFF
--- a/src/Driver/RemoteClient.php
+++ b/src/Driver/RemoteClient.php
@@ -313,10 +313,12 @@ final class RemoteClient implements Client
 
         if ($close) {
             $this->status |= self::CLOSED_WR;
-            return $this->socket->end($data);
+            $this->socket->end($data);
+            return Future::complete();
         }
 
-        return $this->socket->write($data);
+        $this->socket->write($data);
+        return Future::complete();
     }
 
     /**
--- a/src/Driver/RemoteClient.php
+++ b/src/Driver/RemoteClient.php
@@ -313,7 +313,8 @@ final class RemoteClient implements Client
 
         if ($close) {
             $this->status |= self::CLOSED_WR;
-            $this->socket->end($data);
+            $this->socket->write($data);
+            $this->socket->end();
             return Future::complete();
         }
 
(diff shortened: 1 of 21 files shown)


src/Driver/RemoteClient.php
```diff
@@ -313,10 +313,12 @@ private function write(string $data, bool $close = false): Future
 
         if ($close) {
             $this->status |= self::CLOSED_WR;
-            return $this->socket->end($data);
+            $this->socket->end($data);
```
kelunik, line 316: `end` does no longer accept optional data, so this needs to be converted into separate `write` and `end` calls.

src/Driver/UpgradedSocket.php
```diff
@@ -48,14 +51,14 @@ public function __destruct()
...
 
-    public function write(string $data): Future
+    public function write(string $data): void
     {
-        return $this->socket->write($data);
+        $this->socket->write($data);
     }
 
-    public function end(string $finalData = ""): Future
+    public function end(string $finalData = ""): void
```
kelunik, line 59: The optional argument has been removed.

src/HttpServer.php
```diff
@@ -267,6 +267,7 @@ public function start(): void
             try {
                 $this->stop(self::DEFAULT_SHUTDOWN_TIMEOUT);
             } finally {
+                var_dump($exceptions);
```
kelunik, line 270: Left over that should be removed.

kelunik:
Looks pretty good, thanks! I just had some minor comments.

src/Driver/UpgradedSocket.php
iggyvolz: I wasn't sure whether I should be changing the `Future`s to `void`s anywhere else - only hit this one because the type changed in Socket.

kelunik (approved):

Issue #324, opened 2021-11-14
Compression is incompatible with streaming responses


Nek-:
I tried to make a super-simple SSE server (code below). It was just not working until I disabled the compression. Since I assume streamed responses should work with compression enabled, I open an issue about it.

```php
use Amp\ByteStream\ResourceOutputStream;
use Amp\Http\Server\HttpServer;
use Amp\Http\Server\RequestHandler\CallableRequestHandler;
use Amp\Http\Server\Response;
use Amp\Http\Status;
use Amp\Log\ConsoleFormatter;
use Amp\Log\StreamHandler;
use Amp\Socket;
use Amp\Http\Server\Request;
use Monolog\Logger;
use Amp\ByteStream\IteratorStream;
use Amp\Producer;


function newEvent() {
    $id = mt_rand(1, 1000);
    return ['id' => $id, 'title' => 'title ' . $id, 'content' => 'content ' . $id];
}

$events = [
    newEvent(),
    newEvent(),
];

Amp\Loop::run(static function () use (&$events) {
    $cert = new Socket\Certificate(__DIR__ . '/../ssl/cert.pem', __DIR__ . '/../ssl/key.pem');

    $context = (new Socket\BindContext)
        ->withTlsContext((new Socket\ServerTlsContext)->withDefaultCertificate($cert));

    $servers = [
        Socket\Server::listen("0.0.0.0:1337"),
        Socket\Server::listen("[::]:1337"),
        Socket\Server::listen("0.0.0.0:1338", $context),
        Socket\Server::listen("[::]:1338", $context),
    ];

    $logHandler = new StreamHandler(new ResourceOutputStream(STDOUT));
    $logHandler->setFormatter(new ConsoleFormatter);
    $logger = new Logger('server');
    $logger->pushHandler($logHandler);

    $server = new HttpServer($servers, new CallableRequestHandler(static function (Request $request) use (&$events) {
        if ($request->getUri()->getPath() === '/') {
            return new Response(
                Status::OK,
                [
                    "content-type" => "text/html; charset=utf-8"
                ],
                <<<FRONT
                <!DOCTYPE html>
                <html lang="en">
                    <head>
                    <title>Yo</title>
                    </head>
                    <body>
                        <h1>Hello World!</h1>
                        <div id="news"></div>
                        <script>
                        //*
                        var news = document.getElementById('news');
                        const evtSource = new EventSource("/sse");
                        evtSource.addEventListener('news', function (event) {
[...]

PR #334, opened 2022-05-29
Use timeouts to avoid long buffering in CompressionMiddleware 

trowski:
This addresses #324 by using an adjustable timeout (100ms by default) when checking the body size to determine if the body should be compressed or when buffering chunks while sending the body. When the timeout expires, any buffered data is sent, regardless of if it meets the minimum chunk size.

For v3, the minimum chunk size param has been removed as such functionality is better within a stream implementation instead of being handled by the middleware. Cancellations also keep the implementation simpler. An untagged commit in byte-stream is required, https://github.com/amphp/byte-stream/commit/092e9316f8b24edd620a1515ee01e487080f16ad. See https://github.com/amphp/http-server/commit/879a0f36504d348846956c08fe1416c338a77262 for the v3 implementation.


src/Middleware/CompressionMiddleware.php
```diff
@@ -49,6 +56,9 @@ public function __construct(
             throw new \Error("The chunk size must be positive");
         }
 
+        if ($bufferTimeout < 1) {
+            throw new \Error("The buffer timeout must be be positive");
```
kelunik, line 60: Double "be"

kelunik (approved):

kelunik:
The v3 implementation also LGTM.

PR #339, opened 2022-08-16
Fix setting buffer timeout in CompressionMiddleware

MBauerDC:
Regarding [this issue](https://github.com/amphp/http-server/issues/324), the merged change added a constructor parameter to the CompressionMiddleware, but did not set the corresponding property in the constructor. This PR rectifies that.


kelunik:
Thanks!

PR #340, opened 2022-12-11
Add HttpServerSupervisor interface

### LINKED ISSUE #338: RFC: Add start() and stop() methods to HttpServer interface (v3+)
I'm digging into the 3.0 beta and from a testing perspective I appreciate the HttpServer interface. It allows to more easily write unit tests in addition to integration tests. However, this interface does not include methods to start and stop the server, limiting its usefulness from a testing perspective.

Are the maintainers open to including the existing start() and stop() methods on the `SocketHttpServer` in the `HttpServer` interface?

trowski:
Closes #338. @cspray and @kelunik what do you think?

### DIFF
--- a/src/HttpServer.php
+++ b/src/HttpServer.php
@@ -2,10 +2,12 @@
 
 namespace Amp\Http\Server;
 
-use Amp\Socket\SocketServer;
-
 interface HttpServer
 {
+    public function start(RequestHandler $requestHandler, ErrorHandler $errorHandler): void;
+
+    public function stop(): void;
+
     /**
      * @param \Closure(HttpServer):void $onStart
      */
@@ -15,11 +17,6 @@ interface HttpServer
      * @param \Closure(HttpServer):void $onStop
      */
     public function onStop(\Closure $onStop): void;
-
-    /**
-     * @return list<SocketServer>
-     */
-    public function getServers(): array;
 
     public function getStatus(): HttpServerStatus;
 }


kelunik (approved):

trowski:
Should the `HttpServer` interface be renamed to `HttpServerObserver`? Maybe this is more appropriate if we are going to have two interfaces.

kelunik:
Maybe `HttpServerListener`?

trowski:
Thinking about it, either `HttpServerObserver` or `HttpServerListener` seems wrong. This is the event emitter, so `HttpServerEventEmitter` would be better. However that doesn't really cover `getServers()` either, so…

```php
interface HttpServer extends HttpServerEventEmitter, HttpServerSupervisor {
    public function getServers(): array;
}

interface HttpServerEventEmitter {
    public function onStart(...): void;
    public function onStop(...): void;
}
```

Thoughts?

cspray:
@trowski that looks really good to me and should cover my use case quite well. I like splitting up the interfaces.

trowski:
I removed `getServers()` from `HttpServer`, as requiring a list of `SocketServer` seems specific to the socket implementation and probably wouldn't apply to any implementer. I only found it used in tests where a `SocketHttpServer` instance was being used directly.

src/HttpServer.php
```diff
@@ -4,22 +4,10 @@
...
-     * @param \Closure(HttpServer):void $onStop
-     */
-    public function onStop(\Closure $onStop): void;
-
     /**
      * @return list<SocketServer>
      */
     public function getServers(): array;
-
-    public function getStatus(): HttpServerStatus;
```
kelunik, line 24: Where did getStatus go?

trowski: It's on `HttpServerSupervisor` only.

PR #342, opened 2023-01-24
Fix: avoid duplicated data for streamed response

### LINKED ISSUE #341: Duplicate data sent by compression middleware when no contentLength
I found an issue in the CompressionMiddleware in the case of streaming a response with no content length.

The following line is filling the `$bodyBuffer` variable:

https://github.com/amphp/http-server/blob/850f1b22ad0f3459c6ef6e89e897c8abf2b18b4c/src/Middleware/CompressionMiddleware.php#L141

While it's also done some lines after... Without cleaning the original var.

https://github.com/amphp/http-server/blob/850f1b22ad0f3459c6ef6e89e897c8abf2b18b4c/src/Middleware/CompressionMiddleware.php#L195-L197

I don't understand why it's done this way so I'm not sure I can provide a fix (something tells me it's not just stupid but has a purpose that I don't understand).

Nek-:
In the case of a streamed response, the content-length is null, so the bodyBuffer is filled at some point.

The problem is that some lines after it's filled again with the same data, lead to duplication (only for the first bytes).

This fix has been suggested by @trowski in #341 .

This PR fixes #341

### DIFF
--- a/src/Middleware/CompressionMiddleware.php
+++ b/src/Middleware/CompressionMiddleware.php
@@ -141,6 +141,7 @@ final class CompressionMiddleware implements Middleware
                     $bodyBuffer .= $chunk = yield Promise\timeout($promise, \max(1, $expiration - Loop::now()));
 
                     if (isset($bodyBuffer[$this->minimumLength])) {
+                        $promise = $body->read();
                         break;
                     }
 


trowski:
Thanks!

PR #345, opened 2023-02-19
fix PHP Fatal error:  Declaration

lav45:


trowski:
I updated this some time ago in a branch, which I've now pushed to the `3.x` branch.

I'd recommend using only the tagged betas to avoid dependency conflicts.

Thanks for taking the time to make a PR!

PR #347, opened 2023-04-02
ForwardedMiddleware

trowski:


composer.json
```diff
@@ -34,9 +34,9 @@
         "amphp/byte-stream": "^2",
         "amphp/cache": "^2",
         "amphp/hpack": "^3",
-        "amphp/http": "^2",
+        "amphp/http": "dev-2.x-split-header as 2.0",
         "amphp/pipeline": "^1",
-        "amphp/socket": "^2",
+        "amphp/socket": "2.x-dev",
```
trowski: Changes here will be reverted once tagged

examples/hello-world.php
trowski, line 1: Will revert these changes, was for demo purposes only.

src/Middleware/ForwardedForMiddleware.php
```diff
@@ -0,0 +1,126 @@
...
+        if ($clientAddress instanceof InternetAddress && $this->isTrusted($clientAddress)) {
+            $request->setAttribute(self::class, $this->getForwarded($request));
+        } else {
+            $request->setAttribute(self::class, []);
+        }
+
+        return $requestHandler->handleRequest($request);
+    }
+
+    private function isTrusted(InternetAddress $address): bool
```
kelunik:
```suggestion
    private function isTrustedProxy(InternetAddress $address): bool
```

src/Middleware/ForwardedForMiddleware.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * @return list<ForwardedFor>
+     */
+    private function getForwarded(Request $request): array
+    {
+        $maps = Http\parseMultipleHeaderFields($request, 'forwarded');
+        if (!$maps) {
+            $forwardedFor = Http\splitHeader($request, 'x-forwarded-for');
```
kelunik: We should probably make the headers we evaulate configurable? Otherwise, if one isn't set / reset by the proxy, there'll be a vulnerability.

src/Middleware/ForwardedForMiddleware.php
```diff
@@ -0,0 +1,126 @@
...
+        if (\str_contains($ip, ':')) {
+            $ip = \trim($ip, '[]');
+        }
+
+        try {
+            return new InternetAddress($ip, $port);
+        } catch (SocketException) {
+            return null;
+        }
+    }
```
kelunik: We should probably add that to `amphp/socket`?

src/SocketHttpServer.php
```diff
@@ -195,17 +215,7 @@ public function start(RequestHandler $requestHandler, ErrorHandler $errorHandler
             $this->logger->warning("The 'xdebug' extension is loaded, which has a major impact on performance.");
         }
 
-        if ($this->compressionMiddleware) {
-            if (!\extension_loaded('zlib')) {
-                $this->logger->warning(
-                    "The zlib extension is not loaded which prevents using compression. " .
-                    "Either activate the zlib extension or disable compression in the server's options."
-                );
```
kelunik, line 203: Where should we put this warning now?

src/Middleware/ForwardedForMiddleware.php
```diff
@@ -0,0 +1,126 @@
...
+
+        $this->trustedIps = new LocalCache($cacheSize);
+    }
+
+    public function handleRequest(Request $request, RequestHandler $requestHandler): Response
+    {
+        $clientAddress = $request->getClient()->getRemoteAddress();
+
+        if ($clientAddress instanceof InternetAddress && $this->isTrusted($clientAddress)) {
+            $request->setAttribute(self::class, $this->getForwarded($request));
```
kelunik: If the request comes from a trusted proxy, we can usually trust its added value in the header, but not _all_ values in the header.

trowski: How should we approach this? Only provide the first array element or wrap them into another object, perhaps providing a `getTrusted(): ForwardedFor` and `getUntrusted(): list<ForwardedFor>`?

src/SocketHttpServer.php
trowski: Ah, that should have been moved to the static constructors.

src/Middleware/ForwardedForMiddleware.php
trowski: That might be handy. We could maybe avoid the exception then too on invalid cases.

src/SocketHttpServer.php
trowski: I've swapped the $compressionMiddleware parameter for a boolean $enableCompression, which shows the warning now instead of throwing from the `CompressionMiddleware` constructor.

src/Middleware/ForwardedForMiddleware.php
trowski: Added `ForwardedForHeaderType` enum and made it a required constructor parameter and required parameter on `createForBehindProxy`.

src/Middleware/ForwardedForMiddleware.php
kelunik: Probably only one element, yes. I think it'll have to be the last (or if that is another trusted proxy, the second last). Providing the others in a wrapping object sounds fine, too. They can be some indicator for humans in attack analysis, but should never be used for security decisions.

src/Middleware/ForwardedFor.php
```diff
@@ -0,0 +1,28 @@
...
+
+    public function getAddress(): InternetAddress
+    {
+        return $this->address;
+    }
+
+    public function getFieldValue(string $key): ?string
+    {
+        return $this->fields[$key] ?? null;
+    }
```
kelunik: The header is well specified, no? So we can add proper getters?

src/Middleware/ForwardedForMiddleware.php
```diff
@@ -0,0 +1,137 @@
...
+
+            $forwardedFor[] = new ForwardedFor($address, $map);
+        }
+
+        return $forwardedFor;
+    }
+
+    private function getForwardedFor(Request $request): array
+    {
+        $forwardedFor = Http\splitHeader($request, 'x-forwarded-for');
```
kelunik: We should test what Nginx / Apache does if the client provides a header value of `x-forwared-for: "`.

src/Middleware/ForwardedForMiddleware.php
trowski: Yes, last element which is not itself a trusted proxy. I think then it may be best to an object where only that IP is provided. The header value itself is always available in the request for logging, etc. if someone desires.

src/Middleware/ForwardedFor.php
trowski: Yes, I should have mentioned I intended to expand upon this object. I was being a bit lazy for the PoC :wink:

src/Middleware/ForwardedForMiddleware.php
trowski: They append the IP of the client to the list, assuming the client is itself also a proxy. Assuming that client isn't in the list of trusted proxies it doesn't really matter because we'll only use the IP appended by nginx / apache.

kelunik: No, I mean.. If the proxy just appends, quote handling will break and everything will just return null.

src/Middleware/ForwardedForMiddleware.php
trowski: Updated to return the last element which itself is not a trusted proxy.

src/Middleware/ForwardedForMiddleware.php
trowski: If the header is invalid, the attribute set by `ForwardedForMiddleware` will be null. At that point I guess it's up to the user to decide what to do with a request that failed to parse the header.

src/Middleware/ForwardedFor.php
trowski: Updated with proper getters.

examples/hello-world.php
```diff
@@ -45,15 +45,10 @@
...
-        $forwardedFor = \implode(', ', \array_map(
-            fn (ForwardedFor $ff) => $ff->getAddress(),
-            $request->getAttribute(ForwardedForMiddleware::class),
-        ));
-
         return new Response(
             status: HttpStatus::OK,
             headers: ["content-type" => "text/plain; charset=utf-8"],
-            body: "Hello, World! " . $forwardedFor,
+            body: "Hello, World! " . $request->getAttribute(ForwardedForMiddleware::class)?->getFor(),
```
kelunik: Should we use `ForwardedForMiddleware::class` or `ForwardedFor::class` for the attribute name?

trowski: `ForwardedFor::class` seems better. Updated.

src/Middleware/ForwardedForMiddleware.php
```diff
@@ -0,0 +1,141 @@
...
+use Amp\Cache\LocalCache;
+use Amp\Http;
+use Amp\Http\Server\Middleware;
+use Amp\Http\Server\Request;
+use Amp\Http\Server\RequestHandler;
+use Amp\Http\Server\Response;
+use Amp\Socket\CidrMatcher;
+use Amp\Socket\InternetAddress;
+
+final class ForwardedForMiddleware implements Middleware
```
kelunik: We can name this also `ForwardedMiddleware`.

src/Middleware/ForwardedForMiddleware.php
```diff
@@ -0,0 +1,141 @@
...
+            return $ip . ':0';
+        }, $forwardedFor);
+
+        /** @var InternetAddress[] $forwardedFor */
+        $forwardedFor = \array_filter(\array_map(InternetAddress::tryFromString(...), $forwardedFor));
+        foreach (\array_reverse($forwardedFor) as $for) {
+            if (!$this->isTrustedProxy($for)) {
+                return new ForwardedFor(
+                    for: $for,
+                    host: $request->getHeader('x-forwarded-host'),
```
kelunik: Often, the `host` header is directly preserved.

PR #352, opened 2023-08-20
Add HttpErrorException

kelunik:
This allows easier error handling and saves us e.g. from having a `FormMiddleware` in `http-server-form-parser` just for the purpose of delegating to the error handler on invalid form submissions.


trowski (approved):

trowski:
Just need to add strict types to the new file.

PR #353, opened 2023-09-01
Change PSR-7 to PSR-3 in middleware example

cwhite92:
I think this is meant to be PSR-3 and not PSR-7, as PSR-3 is Logger interfaces and PSR-7 is HTTP message interfaces. Confused me for a quick sec.


kelunik (approved):

kelunik:
Of course, thank you!

PR #372, opened 2024-10-21
Add `_` to host validation

maximal:
Local hosts can contain underscores. For example in Docker contexts (application/services names).


kelunik (commented):
Hmmm... underscores are not valid in hostnames or domain labels as far as I know.

Any reason why you try to use underscores instead of dashes?

PR #373, opened 2024-10-28
Fix method property in logger context

luzrain:
Fix method property in logger context in AbstractHttpDriver::handleInternalServerError(). Should be method as string.


kelunik (approved):

kelunik:
Thanks!

Issue #374, opened 2024-10-28
ErrorHandler interface limitations.


luzrain:
[ErrorHandler](https://github.com/amphp/http-server/blob/3.x/src/ErrorHandler.php)  is pretty useless now, it does not give me the ability to implement my own Error handler.
By Error handler I mean my own logic that takes an exception and request as input arguments.

I would like to complete replace all of this default logic that hardcoded in AbstractHttpDriver now:
https://github.com/amphp/http-server/blob/26001b1ffac11db30c627ed4daba2088951b0ffa/src/Driver/Internal/AbstractHttpDriver.php#L90
with my own implementation and log exceptions by myself as I need, but now I can't.
I can't event intercept an exception to get access to the stack trace. Current ErrorHandler only gives access to exception code. Having only code in Error handler is absolutely not enough.

Here is an example how from my point of view [ErrorHandler](https://github.com/amphp/http-server/blob/3.x/src/ErrorHandler.php) interface should look like.

```php
interface ErrorHandler
{
    public function handleError(\Throwable $throwable, ?Request $request = null): Response;
}
```

Then the default ErrorHandler implementation could do the default logging that is now hardcoded in AbstractHttpDriver and generate Response for the browser, and I could completely replace the implementation having access to the exception itself as I need.

Are you interested in such changes? If so, I could work on it. But of course it will not be backwards compatible.
What do you think about it?

kelunik:
Sounds good to me. We should probably name the new interface `ExceptionHandler`, so we can avoid the naming conflict. If an `ErrorHandler` is passed, we can likely automatically convert it into an `ExceptionHandler` and add the logging logic in there.

trowski:
`ErrorHandler` was not designed for exception handling. It was designed to provide a response to bad or malformed requests, which is why it does not require an exception to be passed.

`AbstractHttpDriver::handleInternalServerError()` also was not meant to be the general purpose exception logger for an application. Rather it should only be hit during development or as the last possible point to catch an unexpected exception from the application rather than crashing the entire server. If we allow overriding the code handling these exceptions, we'll need the existing exception handler to catch exceptions from that user-defined handler.

The request handler provided to the HTTP server should be catching all exceptions and logging them as you desire. I don't see a reason to provide this functionality at a lower level, separate from the request handler.

dam2k:
It should also be noted that Amp\Http\HttpStatus\DefaultErrorHandler make use of a blocking function ( file_get_contents() ) to read on resources/error.html

[...]

PR #375, opened 2024-11-09
Add ExceptionHandler interface and middleware

trowski:
This adds a new interface `ExceptionHandler` discussed in #374 which is designed to handle uncaught exceptions from a `RequestHandler` as part of the middleware stack using `ExceptionHandlerMiddleware`. The interface specifies a single method which takes the uncaught exception and associated `Request`, returning a `Response`.

There is a default implementation provided by `DefaultExceptionHandler`, which logs the uncaught exception to the `LoggerInterface` instance provided, then returns a response generated by a delegate `ErrorHandler`.

I've added a parameter to the static constructors of `SocketHttpServer` for better discovery of the interface, and will add docs to the readme if merged.


src/DefaultExceptionHandler.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+final class DefaultExceptionHandler implements ExceptionHandler
+{
+    public function __construct(
+        private readonly ErrorHandler $errorHandler,
+        private readonly PsrLogger $logger,
+    ) {
+    }
+
+    public function handleException(\Throwable $exception, Request $request): Response
```
kelunik: I'd put the `Request` as first parameter, I think.

src/Driver/Internal/AbstractHttpDriver.php
```diff
@@ -80,67 +80,26 @@ final protected function handleRequest(Request $request): void
...
-                'method' => $method,
-                'uri' => $uri,
-                'protocolVersion' => $protocolVersion,
-                'local' => $local,
-                'remote' => $remote,
-            ],
-        );
-
-        return $this->handleError($status, null, $request);
+        $this->exceptionHandler ??= new DefaultExceptionHandler($this->errorHandler, $this->logger);
```
kelunik: Is it really worth making this nullable instead of eagerly initialized?

src/Driver/Internal/AbstractHttpDriver.php
```diff
@@ -80,67 +80,26 @@ final protected function handleRequest(Request $request): void
...
-                    $exception::class,
-                    $this->errorHandler::class,
-                ),
-                ['exception' => $exception],
-            );
-
-            // The default error handler will never throw, otherwise there's a bug
-            return self::getDefaultErrorHandler()->handleError($status, null, $request);
-        }
+        return $this->errorHandler->handleError($status, $reason, $request);
```
kelunik: Shouldn't we inline this method then?

src/Driver/Internal/AbstractHttpDriver.php
trowski: No, probably not. Changed.

src/DefaultExceptionHandler.php
trowski: Swapped.

src/Driver/Internal/AbstractHttpDriver.php
trowski: Might as well, yes. I also inlined the call to the exception handler and moved the comment about not being designed as the general-purpose catch block to call site.

src/DefaultExceptionHandler.php
```diff
@@ -0,0 +1,54 @@
...
+        $client = $request->getClient();
+        $method = $request->getMethod();
+        $uri = (string) $request->getUri();
+        $protocolVersion = $request->getProtocolVersion();
+        $local = $client->getLocalAddress()->toString();
+        $remote = $client->getRemoteAddress()->toString();
+
+        $this->logger->error(
+            \sprintf(
+                "Unexpected %s with message '%s' thrown from %s:%d when handling request: %s %s HTTP/%s %s on %s",
```
kelunik, line 31: I'd somehow expect a separator between `%s %s` between the HTTP version and the remote address, but not sure how exactly.

kelunik: Shouldn't we use PSR placehodlers in the message?

kelunik (approved):

PR #386, opened 2026-04-04
Fix http/2 ping flood behaviour

kafkiansky:
Hi. I’m building a [gRPC](https://github.com/thesis-php/grpc) server with long-lived client/server/bidirectional streams.
In such streams it is normal to receive many HTTP/2 PING frames without new HEADERS, because subsequent gRPC messages are sent as DATA on an already open stream.

In `amphp/http-server`, the ping counter was effectively reset on new request headers, but not on incoming stream data. That caused active gRPC streams to be misclassified as ping abuse, eventually triggering GOAWAY (Too many pings).

My fix is to reset the ping counter when DATA is received on an active stream. Inbound DATA is valid application activity, so these pings are not an idle ping flood.

### DIFF
--- a/src/Driver/Http2Driver.php
+++ b/src/Driver/Http2Driver.php
@@ -1169,6 +1169,8 @@ final class Http2Driver extends AbstractHttpDriver implements Http2Processor
 
     public function handleData(int $streamId, string $data): void
     {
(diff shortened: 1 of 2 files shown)


trowski:
Thanks!

